PR #134, opened 2020-11-08
remove actor-classes template, as it doesn't exist

genmaster:
to avoid file not found errors in js console


moo-man:
Thanks but this is already done in the develop branch

PR #143, opened 2020-11-09
Added Lores (Witches and Torture)

ericbrooke:
Do not Merge as the last "_id" is not unique, hopefully this makes it easier for you.

Lores are for WitchHunter Class


moo-man:
Hi, thanks for the PR, however I'm not keen on having a huge list of specializations. What I've included in the Skill compendium pack, both basic and core, is just the example specializations given in the skill, and an empty specialization to be filled for anything else. It's impossible to include everything (as skills are infinite, essentially), so I'd rather not try.

ericbrooke:
I respect that this is your space. I wonder if we could extend the list to only include things that are in the first career ladder. Make on-boarding much easier and in many cases you already have them in there i.e. career descriptions just not in Skills and trappings. I am happy to help

moo-man:
I'm curious what part of onboarding needs improvements. I've created functions that should cover this problem.

The `findSkill` function will take a skill name, (say `Lore (Witches)`), if it can't find `Lore (Witches)` verbatim, it will try to find `Lore ()`, when it does so, it will rename the skill to `Lore (Witches)` and return it. This is what's used in character creation and when adding untrained career skills to the actor.

ericbrooke:
You are right. I tried this once and it did not work i.e. for the WitchHunter and Lore(Torture). I just re created the character from scratch and this time it worked. To get Lore(Witches) I had to use the second career level put on the character state current and then goto skills and click on Lore(Witches) and then remove the Career Ranking

moo-man:
Right, because Lore (Witches) is a second tier skill, I guess I'm just confused on what you're trying to do? You are trying to get Lore (Witches) at character creation? Because that's not how it would work by the character creation rules.

Issue #136, opened 2020-11-08
Define a default CSS text color on an application level


Moerill:
Coming from this issue:
https://github.com/Moerill/fvtt-pointer/issues/6

![image](https://user-images.githubusercontent.com/26152015/98462651-21d05d00-21b6-11eb-8c36-08d76571c924.png)

What is happening?
I (and probably many other mods with custom apps) do not define a text color and background, so the theme can take care of that. But most text colors you define in your system theme do only get applied on more specific classes/parts of an application and not on the higher level. So if someone doesn't follow that class schema, it will just be default FVTTs dark gray, which is nigh unreadable.
Proposed solution:
add a css rule, that overrides the default fvtt app text color rule ``.window-app .window-content { color: #191813; }`` with something that is readable on most backgrounds of default apps for your system.

valberg:
I'm having this same issue with many help pages and such defined by modules.

PR #301, opened 2021-01-06
Make default font color in window content light

valberg:
Basically doing what Moerill is suggesting in #136.


moo-man:
I know I've tried this and it's caused other conflicts. Will have to go through various windows to ensure it's all good.

valberg:
Ah, I could try to go through all windows, but I guess you have a better overview of possible windows that would be affected.

moo-man:
I can handle this myself just documenting here

- [ ] Text Editor Tables
- [ ] File Browser

PR #331, opened 2021-01-23
Update utility-wfrp4e.js

Cododoc:
Fix get actor by id.


moo-man:
Appreciate the PR, but I'm assuming this is to fix traits in opposed rolls. 

This isn't an issue with `getSpeaker()`, this is an issue with how `speaker` is constructed in `setupTrait()`, which has been fixed in `develop`.

PR #673, opened 2021-06-27
few fixes (not 100% tested though)

silentmark:
Hi, 

few suggestions for fixes and tweaks after refactor. Great job btw. I hope that those will help, feel free to reject the PR and apply only part of those manually, if not all of them are relevant.

### DIFF
--- a/modules/actor/actor-wfrp4e.js
+++ b/modules/actor/actor-wfrp4e.js
@@ -307,7 +307,7 @@ export default class ActorWfrp4e extends Actor {
     let size;
     let trait = this.has(game.i18n.localize("NAME.Size"))
     if (trait)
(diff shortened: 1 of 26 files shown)


moo-man:
Thanks, I'll use this as a guide, but there's definitely some changes I want to make

For instance, the `skills` getter in 
`let skill = this.skills.find(i => i.name == item.data.test.value)`

I would prefer to just do `this.getItemTypes("skill).find(s => s.name == test.data.test.value)`

And since I want to use object instances as much as possible, and utilize their getters, it'd probably be `test.test.value`.

Additionally, `extra.reroll` I think should be `context.reroll`, but this is just from a cursory glance.

Feel free to submit further changes, I'll submit changes to the PR as well just to keep track of them.

PR #741, opened 2021-07-26
few localization improvments

silentmark:


modules/system/rolls/cast-test.js
```diff
@@ -74,6 +74,7 @@ export default class CastTest extends TestWFRP {
       // Miscast on fumble
       if (this.result.roll % 11 == 0 || this.result.roll == 100) {
         this.result.color_red = true;
+        this.result.other.push(game.i18n.localize("CHAT.FumbleMiscast"))
```
moo-man, line 77: I'm not very keen on using `other` for this, which I try to reserve for descriptions of further effects, not explanations of what's already on the card, if that makes sense. I might put this in a tooltip when you hover over miscast.

silentmark: sure, then i will remove it. personally i found that useful, but maybe there are better ways for this

moo-man: No need to remove it, I can just make the change to this PR.

PR #836, opened 2021-09-11
Added Spanish translation

Txus5012:
Hi.

Thanks again for your hard work. I wanted to contribute with a Spanish translation I made based in the Spanish Core Rulebook by Devir.

Almost everything in the lang file is translated, except for a few things that would break some automations if translated:

- Characteristics: The full names aren't translated. The abbrevs are, as far as I've tested they didn't break anything.
Bonuses aren't translated either.
- Skills, Talents and Traits.
- Conditions: They would work, but the links would show 'undefined'.
- Weapon Groups: Because they have to match the Skills.
- Money (GC, SS and CP).

Hit Locations are translated, the only issue is that the opposed test shows _Bodypart_ (Remapped). I think we can live with that.

Also, I made the 'Blessed By' label and the Scatter table slightly wider. Now spanish words fit and english users shouldn't notice it.


moo-man:
Hi, thanks, but generally I recommend people create a separate module for translations, and I'm pretty sure one already exists for Spanish

Txus5012:
I see. I'll try to learn a bit about how to make modules.
I haven't seen a Spanish translation.

Txus5012:
Oops, I didn't know this would be updated. I'm very new to this.
How can I cancel this pull request?

moo-man:
Just "close" at the bottom, which I'll do with this comment

You can see how the french translation is done here https://gitlab.com/LeRatierBretonnien/foundryvtt-wh4-lang-fr-fr

PR #844, opened 2021-09-16
Added Overcast Reset

Txus5012:
Hi. While creating the translation file I noticed that the Overcast Reset line was missing. Here is the fix.

By the way, I fixed issues 839 and 689 (journal fixes. Just wanted to see if I could do it). But they are in the core module.
Do you want the fixes? If so, how can I send them to you?


moo-man:
Thanks, but there's no need to send me module fixes, I will handle them

PR #944, opened 2021-12-09
Add error messages advancement functions to prevent negative xp

Khakimonk:
I've added some small code to prevent xp going into negative, also added a i18n message and ui errors to display on click. It's not my most DRY code, but I didn't want to go adding a function in the sheet. I could add it in the utility class if that works better for you?

```
static checkValidAdvancement(total, spent, action, item) {
  if(total - spent < 0) {
     ui.notifications.error(game.i18n.format("ACTOR.AdvancementError", { action: action, item: item }));
     return false;
  } else {
    return true;
  }
}
```

for example

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -228,6 +228,7 @@
     "ACTOR.AddBasicSkills" : "Add Basic Skills",
     "ACTOR.ImportStatBlock" : "Import Stat Block",
     "ACTOR.SpentExp" : "Spent {amount} Experience on {reason}",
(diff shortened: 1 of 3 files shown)


moo-man:
Thanks for this, yeah I think the utility function is a good idea and should make it a bit cleaner, and I would probably `throw new Error` inside of it so you don't have to check whether it returned true or false.

Khakimonk:
Ok, so added the utility function and wrapped in a try/catch then just display the error. I chose not to post it to the console as it's a generated error not a code error.

moo-man:
Great, thanks!

Issue #992, opened 2022-01-13
[WFRP4e System] Player Actor Opposed Test Error Using Fortune


hobeika:
**Describe the bug**
During an encounter, when a player is defending against a successful attack (positive SL) with an opposed test, if the player fails the test and tries to use a fortune point (or a dark deal), an error arises.
The test is indeed rerolled, but the damages are not computed and displayed again and the following error message can be seen in the console:
​
 ```
Foundry VTT | Created Combat with id [eN3EEnwHDrtJ36Q9]
​ Foundry VTT | Created 2 Combatant documents with ids: [TUQWrD4ODaH0LlaI,RLGPxAY4P7dZ7YT9] in parent Combat [eN3EEnwHDrtJ36Q9]
​ Foundry VTT | Created 2 ChatMessage documents with ids: [zEcyFPxUxsf7fAQL,LZ93HHw6TFO9NsVL]
​ Foundry VTT | Retrieved and compiled template templates/dice/tooltip.html
​ Foundry VTT | Retrieved and compiled template templates/dice/tooltip.html
​ Foundry VTT | Retrieved and compiled template templates/dice/roll.html
​ Foundry VTT | Retrieved and compiled template templates/dice/roll.html
​ Foundry VTT | Retrieved and compiled template systems/wfrp4e/templates/chat/chat-message.html
​ Foundry VTT | Loading audio element - sounds/dice.wav
​ Foundry VTT | Retrieved and compiled template systems/wfrp4e/templates/chat/chat-message.html
​ Foundry VTT | Loading audio buffer - sounds/dice.wav
​ Foundry VTT | Constructing audio buffer source node - sounds/dice.wav
​ Foundry VTT | Created ChatMessage with id [LAHe3syOvRwCh55E]
​ Foundry VTT | Created ChatMessage with id [Umyp9iHwtSrAfbog]
​ Foundry VTT | Loading audio element - null
null:1 Failed to load resource: the server responded with a status of 404 (Not Found)
foundry.js:62998 Foundry VTT | Failed to load audio node: Event
_createNode @ foundry.js:62998
​ Foundry VTT | Created ChatMessage with id [lk9vFH0j5vUgsNiy]
​ Foundry VTT | Loading audio element - null
null:1 Failed to load resource: the server responded with a status of 404 (Not Found)
foundry.js:62998 Foundry VTT | Failed to load audio node: Event
_createNode @ foundry.js:62998
[...]

PR #1000, opened 2022-01-21
Ref: #992 fix the insertion of the ref to the attacker message

hobeika:
The message.data.flags.data.attackerMessage returns an array (with one element) and not the final string with the message id.
This ends up failing to find the message in the map and the attackerMessage stays as undefined.

### DIFF
--- a/modules/system/opposed-wfrp4e.js
+++ b/modules/system/opposed-wfrp4e.js
@@ -127,7 +127,7 @@ export default class OpposedWFRP {
       this.defenderMessage.update({ "flags.data.attackerMessage": this.attackerMessage.id });
     }
     else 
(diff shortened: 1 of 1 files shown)


hobeika:
As I am not a JS developer I'm happy to have feedback on the fix itself. There might be more elegant ways to do this.

hobeika:
After further review the fix is incomplete.
This works on the player side but not the GM side anymore.
For some reason the array is different on the GM side and the Player side.

moo-man:
It might simply be time to refactor this entire process to be cleaner.

hobeika:
Not the best fix I guess but now it works both as a GM and as a player.

moo-man:
I appreciate the contribution but I've entirely reworked the opposed test system, which should resolve this issue.

PR #1024, opened 2022-02-06
Sidebar Macro Icon support added

Sasmira:
Add support for Sidebar Macro module 

Without module 

With module


css/wfrp4e.css
```diff
@@ -826,7 +826,7 @@
 
   #sidebar.collapsed {
     width: 50px !important;
-    height: 364px !important;
+    height: 400px !important;
```
moo-man, line 829: Wouldn't this apply even if the module is not active?

Sasmira: without the height extension, we lose the last icon :D

moo-man: Right, but with the height extension, what does it look like without the module?

Sasmira: Hoo sorry ! on my first message, you have the 2 first screens without the module but already with the height extension actived

moo-man: Ah sorry I misunderstood, thanks.

Sasmira: No problem I don't speak a lot too :D

PR #1036, opened 2022-02-14
fix overlay config, the values appeared under the rings

Sasmira:
Now with the fix


Sasmira:
2  fix :
- Combat Carousel Initiative button
- the button on the right bar didn't appear
- fix overlay config, the values appeared under the rings (it's inside too )

GIF 21-02-2022 03-44-12

PR #1053, opened 2022-02-24
Scrolling Text Variation

Txus5012:
Makes the scrolling text move upwards when the number increases and move downwards when the number decreases. Which is visually (and psychologically) more impacting.


fmunoz-geo:
Nice and clear but too much code copied and pasted. Maybe just create a variable and use it in the direction value.

Or even a something more minimalist like  direction: (change<0) ? 1: 2 ;

Txus5012:
Nice tip. Thanks.

PR #1081, opened 2022-03-10
Sanitize chat message content

hobeika:
When a player enters a single or double quote in the chat message it generates an error.
This has the side effect to reveal whispers to anyone monitoring the errors in the console.

Escaping the string fixes the issue and doesn't seem to break other behaviors.


moo-man:
Thanks, but I've decided to refactor this process away from using Hooks

PR #1135, opened 2022-05-22
Guard against undefined objects during encounters

kbluck:
Hello, learning to develop on this project. Starting simple by eliminating a couple of exceptions that I noticed.

1. When starting a combat encounter without a scene or tokens, undefined object exceptions fire in `CombatHelpers.startTurnChecks()` for the `turn` object and `CombatHelpers.checkEndTurnConditions()` for the `combatant` object. Added tests for undefined to prevent these exceptions, and write a console warning instead in `startTurnChecks()`.

2. Seems like all the sound files are removed from `sounds.txt`, so `WFRP_Audio.PlayContextAudio` by default always throws an undefined object and also causes an HTTP 404. Added test for undefined `sound.file` to prevent both of those in favor of a console warning.

3. Set git to ignore the `wfrp4e.lock` file if you happen to have the system locked in VTT.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,2 @@
-
 debug.log
+wfrp4e.lock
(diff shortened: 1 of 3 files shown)


moo-man:
Thanks!

PR #1137, opened 2022-05-24
Fix for Bulky flaw

Txus5012:
A quick fix for the Bulky flaw, which wasn't adding the extra encumbrance.


fmunoz-geo:
I'm ashamed of myself. It was two lines and botched one. Thanks for the fix.

PR #1152, opened 2022-06-14
Compatibility with Introduce Me

Txus5012:
Adding UI compatibility with [Introduce Me](https://github.com/WBHarry/introduce-me). Now (almost) all of it should be readable in WFRP.


GregorHak:
NICE :)

PR #1163, opened 2022-06-25
Localize

Yasnen:
PR what can be localized with modifications that only affect appearance.

### DIFF
--- a/lang/en.json
+++ b/lang/en.json
@@ -497,6 +497,8 @@
     "the Terrain": "the Terrain",
     "chosen Lore": "chosen Lore",
     "Charging": "Charging",
+    "Players": "Player",
+    "Enemies": "Enemies",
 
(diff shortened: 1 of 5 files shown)


lang/en.json
```diff
@@ -496,6 +496,9 @@
     "Sense": "Sense",
     "the Terrain": "the Terrain",
     "chosen Lore": "chosen Lore",
+    "Charging": "Charging",
+    "Players": "Player",
```
phenomen, line 500: Should be "Players"?

moo-man: @Yasnen Can you fix this? I can then merge

Yasnen:
I don't really understand how to use GitHub, so if I'm wrong, please point it out.

Yasnen:
I don't really understand how to use GitHub, so if I'm wrong, please point it out.

PR #1164, opened 2022-06-25
Fix Group Advantage (combat end, charging)

Jagusti:


modules/hooks/init.js
```diff
@@ -2,6 +2,10 @@ import NameGenWfrp from "../apps/name-gen.js";
 import TravelDistanceWfrp4e from "../apps/travel-distance-wfrp4e.js";
 import HomebrewSettings from "../apps/homebrew-settings.js";
 
+const debouncedReload = foundry.utils.debounce(() => {
+  window.location.reload();
+}, 100);
+
```
moo-man, line 8: There's gotta be a built in way to do this, is there not?

Jagusti: I think for a while this has been the recommended way. 

https://discord.com/channels/170995199584108546/811676497965613117/845037272499748905

moo-man: I'm guessing this obliterates the "Auto Fill Advantage disabled" notification, is there a way to add it after the reload?

Jagusti: I guess mmove this from settings.js to Hooks.once("ready"), adding a check for autoFillAdvantage? 

https://github.com/moo-man/WFRP4e-FoundryVTT/blob/96bf0acdf9c5a208b626837a90065595e1818e43/modules/hooks/settings.js#L29-L34

```
        // Automatically disable Auto Fill Advantage if group advantage is enabled
        if (setting.key == "wfrp4e.useGroupAdvantage" && setting.value && game.user.isGM && game.settings.set("wfrp4e", "autoFillAdvantage", true))
        {
            ui.notifications.notify(game.i18n.localize("AutoFillAdvantageDisabled"), {permanent : true})
            game.settings.set("wfrp4e", "autoFillAdvantage", false)
        }
```

Jagusti: If that works for you I can commit that change to this PR too.

moo-man: I think that works, though it prevents people using both settings if they want, not sure why they would, not a big deal.

PR #1166, opened 2022-06-26
Improve compatibility with Simple Calendar

ZwS:
Replaces simple calendar classic theme with the one that matches system style. Both full and compact views are updated.

For new setups Simple Calendar offers Dark theme by default, which works ok with the WFRP system but looks alien. Classic theme intended to mimic the DnD5e system style, so I chose to replace it to provide system look.

Imgur
Imgur


GregorHak:
Fine, but one thing - It is not easy to see full dark moon
obraz

ZwS:
Hi @GregorHak 
thanks for pointing this out. I've updated style and add small shadow to the icons (similarly as it is done for the note counters)
Imgur

Issue #1091, opened 2022-03-20
[WFRP4e System] A single entry with multiple meanings in a language file corresponds to multiple words in Japanese


moo-man:
The English words "Height" and "Weight" have multiple meanings, corresponding to different words in Japanese.
In the case of "Height", the Japanese word for "身長" is "how tall someone is" and "高さ" is "the distance something is above the ground. The language file does not split items that should be converted into multiple words with multiple meanings, resulting in a strange display.
Please update the language file.

This post was prepared using machine translation by DeepL.
Thanks DeepL.
<br/>**From**: MRyas#3956<br/>foundry: 9.255<br/>wfrp4e: 5.1.1<br/>wfrp4e-core: 2.4.1

PR #1172, opened 2022-07-05
Japanese localize issues

Yasnen:
Due to grammatical differences between English and Japanese, suggested modifications to the language file are desirable.


moo-man:
Fixes #1091

Yasnen:
"Localize xp log" previously PR was embargoed and fixed

PR #1179, opened 2022-07-19
Add test difficulty option for roll dialog bypass

Jagusti:
This PR adds support for passing a test difficulty parameter when setting up a skill, characteristic or other test. As with supplying a test modifier, it is only applied when bypassing the roll dialog. 

If no parameter is passed, system continues to default to challenging or average depending on the default test difficulty module setting. 

Context: I'm overhauling the Make Secret Party Test macro in GM Toolkit to support passing a test difficulty parameter. As it's not supported in test setup by the system, the changes are inconsequential.  I know others have hacked the macro to add the parameter in the past, so it would be nice to add the option out of the box.


moo-man:
Actually this is already supported in `getPrefillData`

You want to set `options.absolute.difficulty` 

Absolute values set the value, instead of modifying.

Jagusti:
Amazing ... cheers muchly!

PR #1201, opened 2022-08-14
Fix datamodel warnings and roll deprecations

Jagusti:
Fixes for errors and warnings I've come across while working on v10 compatibility for GM Toolkit.

### DIFF
--- a/modules/system/audio-wfrp4e.js
+++ b/modules/system/audio-wfrp4e.js
@@ -5,7 +5,7 @@ export default class WFRP_Audio {
         console.warn("wfrp4e | Sound file not found for context: %o", context)
         return
       }
-      WFRP_UTILITY.log(`wfrp4e | Playing Sound: ${sound.file}`)
+      game.wfrp4e.utility.log(`wfrp4e | Playing Sound: ${sound.file}`)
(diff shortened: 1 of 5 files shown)


modules/system/audio-wfrp4e.js
```diff
@@ -5,7 +5,7 @@ export default class WFRP_Audio {
         console.warn("wfrp4e | Sound file not found for context: %o", context)
         return
       }
-      WFRP_UTILITY.log(`wfrp4e | Playing Sound: ${sound.file}`)
+      game.wfrp4e.utility.log(`wfrp4e | Playing Sound: ${sound.file}`)
```
moo-man, line 8: Not that it really makes a difference, but why this change?

Jagusti:
I think this one was because I was getting a WFRP_UTILITY error, as if it
wasn't being imported. I tried adding the import at the head of the file
and it gave me grief for some reason, so I fell back to this to avoid
getting more sidetracked!

I think there's another reference to WFRP_UTILITY at the bottom of this
file, but I've not amended that.  I've tried to only touch those nights of
code that I got an error called for. Hence a lot of the test rolls are not
explicitly async'd like I did basic test.

PR #1210, opened 2022-09-01
WoM overcast rules, channel until ready button, tweaks in spell formula calculations

silentmark:
Hi, 

here are bunch of tweaks that i would like to propose, first one is Winds of Magic overcasting rules for spells. I think this is self-explanatory. 

Second feature is related to spellcasting outside of combat. Not sure how about you, but i have few players that like to waste a lot of time on casting spells from books outside of combat to achieve something and this requires a lot of rolling for channelling. Thats why i came with another button - channel until ready. This will re-roll channelling test using one-time selected options until either CN has been reached or miscast occurred. 

image

Last tweak is related to spell formula calculation. There are few scenarios, where this didn't work for example for Polish translation, as Willpower and Strength are translated to "Siła Woli" and "Siła". One is substring of another which leads to problems. Also, for spellprayer formula - the parsing has been improved to correctly eval formulas with units. 

Two final comments: not sure if this is the right branch to merge (but develop is behind few commits).
I haven't really checked the WoM rules against "other/special" usage of overcast.


moo-man:
Well, this is a bit awkward as I'm not sure what I want to do with WoM mechanics yet, meaning where they are located.

Also, regarding spell formula calculation, this is generally why I prefer translation modules as they can override the function as they need, is that not possible?

silentmark:
i tried to make WoM overcast as separated as possible, exactly because i expected that you might already have something in mind, but since i really like that mechanics i did it for myself and wanted to share. 

Regarding formula - there are two aspects - one is purely polish problem - and that's where the different iterator comes into play for characteristics - it's generic and doesn't break anything and obviously it doesn't have any performance impact or side effects - i wouldn't mind to have it included, but i guess it can be handled by the module. The second part, for computeSpellPrayerFormula - this is i think a generic problem when there are units in place and more complex formula - it simply breaks because on eval - when it expects only numeric formula and gets additional text. Probably it could be written nicer, but i didn't want to go into full expression parsing thing.

moo-man:
Okay sounds good, do you have an example case where the compute method breaks?

silentmark:
So, good example is anything that has any formula in duration/target/aoe except single number of single characteristic/characteristic bonus. For example there is a Breath spell/trait from UG, which has range formula: 20 + Toughness Bonus yards. The original formula will replace this into 20 + X yards. next, inside `prepareOvercastingData` you make simple `unit: range.split(" ")[1]`, and this returns something inaccurate (+ X yards), later, parseInt either returns only the first value, without actual evaluation of the formula, or breaks.

modules/actor/actor-wfrp4e.js
```diff
@@ -3751,14 +3755,14 @@ export default class ActorWfrp4e extends Actor {
...
         token: token?.id || this.token.id,
-        scene: token?.parent.id || this.token.parent.id
+        scene: token?.scene?.id || token?.parent?.id || this?.token?.scene?.id || this?.token?.parent?.id
       }
     }
     else {
       return {
         actor: this.id,
         token: token?.id,
-        scene: token?.parent.id
```
moo-man, line 3761: What was this change for?

modules/actor/sheet/actor-sheet.js
```diff
@@ -513,11 +513,31 @@ export default class ActorSheetWfrp4e extends ActorSheet {
...
+                    if (testObject.result.minormis || testObject.result.majormis || testObject.result.catastrophicmis) {
+                      break;
+                    }
+                    if (testObject.item.cn.SL == testObject.item.cn.value) {
+                      break;
+                    }
+                  } while (true)
+                });
+              }
+            }
```
moo-man, line 540: I think I would prefer this as a checkbox in the channel dialog, what do you think?

modules/system/rolls/cast-test.js
```diff
@@ -188,11 +188,31 @@ export default class CastTest extends TestWFRP {
...
+          this.result.overcast.usage.target.available = this.result.overcast.available >= game.wfrp4e.config.overCastTable.target[0].cost
+        }
+      }
+      if (this.result.overcast.usage.duration) {
+        this.result.overcast.usage.duration.available = this.result.overcast.available >= game.wfrp4e.config.overCastTable.duration[0].cost
+      }
+      if (this.result.overcast.usage.damage) {
+        this.result.overcast.usage.damage.available = this.result.overcast.available >= game.wfrp4e.config.overCastTable.damage[0].cost
+      }
+    }
```
moo-man, line 215: This is making me think it would be cleaner to extend the `CastTest` class into a `WoMCastTest` class that could reduce the  if/else hell. Do you think so?

silentmark: Good idea

modules/actor/actor-wfrp4e.js
silentmark: i will have to re-test this, but when i did those changes, i had some cases, where this code was causing errors - sometimes token had parent - and sometimes scene. and sometimes there was no token for given actor on given scene.

modules/actor/sheet/actor-sheet.js
silentmark: initially i did exatly that, but this would require to make an extra change in channel callback, where i dont have access to testObject and also all those setup<Test>().then(setupData => {}) are not awaited. Therefore it would require to break nice pattern that you have and extra button removes this problem. Also, i think it simply looks better (small chance that someone will do that by mistake).

moo-man: Maybe I'm misunderstanding, but this seems fine? I suppose we disagree because I don't really like the look of the button (especially because it expands the height of them all).

```js
            channel: {
              label: game.i18n.localize("Channel"),
              callback: async btn => {
                let test = await this.actor.setupChannell(spell, options)
                if (test.context.channelUntilSuccess) // Added via a dialog checkbox
                {
                   do {
                    let testData = duplicate(test);
                    let testDuplicate = new test.rollClass(testObject)
                    if (test.result.minormis || test.result.majormis || test.result.catastrophicmis) {
                      break;
                    }
                    if (test.item.cn.SL == test.item.cn.value) {
                      break;
                    }
                  } while (true)
                }
                else 
                   test.roll()
                }
              }
```

Regarding accidental activating this, perhaps we can remove this option if a combat is active?

Also, shouldn't the CN == SL check be the condition for the loop? Such as
`while (test.item.cn.SL < test.item.cn.value)`

silentmark: yeah, but this is excatly the change - any other callback isnt async and setupChannel or any other setupSkill/Weapon etc is never awaited, so this is the pattern i didnt want to break. Maybe they should, but they arent. i.e. those (note, that this example is not accidental, as some of the actions are async and some arent.)

```javascript
  _onWeaponNameClick(ev) {
    ev.preventDefault();
    let itemId = $(ev.currentTarget).parents(".item").attr("data-item-id");
    let weapon = this.actor.items.get(itemId)
    if (weapon) this.actor.setupWeapon(weapon).then(setupData => {
      if (!setupData.abort)
        this.actor.weaponTest(setupData)
    })
  }
  async _onUnarmedClick(ev) {
    ev.preventDefault();
    let unarmed = game.wfrp4e.config.systemItems.unarmed
    this.actor.setupWeapon(unarmed).then(setupData => {
      this.actor.weaponTest(setupData)
    })
  }
``` 

As for do{ } while (true) - i thought that it will be more readable, as we have two break conditions and with miscasts included, the while (condition) would be very long.

moo-man: So this is actually an old pattern from before I gave Tests their own classes. Notice how `basicTest` and `weaponTest` just simply call `test.roll()`

`setupData` shouldn't really be called that anymore because it's a test object when it used to be just its data. 

I didn't change it because I didn't want to break anything at the time. And those functions exist simply for backwards compatibility (mainly with effects)

Regardless the new pattern:
```js
this.actor.setupWeapon(weapon).then(test => test.roll())
```
I think is cleaner than 
```js
this.actor.setupWeapon(unarmed).then(setupData => this.actor.weaponTest(setupData))
```

silentmark: i will experiment on that in this case, i need to dig ia bit more into it, to understand the consequences when the test is awaited and when not.

moo-man: The test? or the setup? 

The test is awaited because rolling is async

The setup is awaited because it returns a Promise that is only resolved when the dialog is submitted, see: https://github.com/moo-man/WFRP4e-FoundryVTT/blob/master/modules/actor/actor-wfrp4e.js#L553-L571

Not sure if that clears up anything but let me know if you have any questions

modules/system/config-wfrp4e.js
```diff
@@ -870,7 +905,6 @@ WFRP4E.conditionDescriptions = {}
 WFRP4E.symptoms = {}
 WFRP4E.symptomDescriptions = {}
 WFRP4E.symptomTreatment = {}
-WFRP4E.conditionDescriptions = {}
```
moo-man, line 873: Why was this removed?

modules/actor/actor-wfrp4e.js
moo-man, line 3761: What's the purpose of this change? Isn't scene the only parent of a token?

modules/actor/actor-wfrp4e.js
```diff
@@ -1087,9 +1088,10 @@ export default class ActorWfrp4e extends Actor {
...
         }
         testData.malignantInfluence = html.find('[name="malignantInfluence"]').is(':checked');
+        testData.channelUntilSuccess = html.find('[name="channelUntilSuccess"]').is(':checked');
         testData.skillSelected = channellSkills[Number(html.find('[name="skillSelected"]').val())];
         testData.cardOptions = cardOptions;
-        if (options.channelUntilSuccess) {
+        if (testData.channelUntilSuccess) {
           return testData;
         } else {
           return new testData.rollClass(testData);
```
moo-man, line 1097: So I probably want to ensure these callbacks **only** return Test objects, not their data.

If you had issues with duplicating the test object, the example I gave *should* work, I remember testing it

```js
let testData = duplicate(test); // Test is a test object, but duplication turns it into just its data
let testDuplicate = new test.rollClass(testData)
```

modules/actor/sheet/actor-sheet.js
```diff
@@ -522,7 +522,8 @@ export default class ActorSheetWfrp4e extends ActorSheet {
                     if (testDuplicate.result.minormis || testDuplicate.result.majormis || testDuplicate.result.catastrophicmis) {
                       break;
                     }
-                    if (test.item.cn.SL == test.item.cn.value) {
+                    let spell = game.actors.get(test.speaker.actor).items.get(test.item)
+                    if (spell.cn.SL == spell.cn.value) {
```
moo-man, line 526: I don't really like this, why was this changed?

modules/hooks/init.js
```diff
@@ -604,14 +604,31 @@ export default function() {
...
     });
 
-    game.settings.register("wfrp4e", "useWoMOvercast", {
-      name: "SETTINGS.useWoMOvercast",
-      hint: "SETTINGS.useWoMOvercastHint",
-      scope: "world",
-      config: true,
-      default: false,
-      type: Boolean
-    });
```
moo-man: Was this a merge error? Feel like you need this

modules/actor/sheet/actor-sheet.js
silentmark: Because when I tested it again, I had only itemid instead of actual item in the object. If you have better idea how to get the spell from its id, I would update it.

modules/hooks/init.js
silentmark: Yes I need this, I think the auto merge with table settings didn't work well

modules/system/config-wfrp4e.js
silentmark: Duplicated

modules/actor/actor-wfrp4e.js
silentmark: Yes, I had problem with that, that was also one of the reasons why I had dedicated button for this instead of checkbox

moo-man:
I've provided the changes and tested them, they seem to work fine, once you confirm I'll merge the PR.

moo-man:
Great, I haven't actually tried out the WoM code yet but once I do I'll merge

moo-man:
@silentmark Please see and review my changes when you can

modules/system/rolls/wom-cast-test.js
```diff
@@ -23,6 +30,9 @@ export default class WomCastTest extends CastTest {
     if (this.result.overcast.usage.damage) {
       this.result.overcast.usage.damage.available = this.result.overcast.available >= game.wfrp4e.config.overCastTable.damage[0].cost
     }
+    if (this.result.overcast.usage.extra) {
```
silentmark, line 33: should this be usage.extra, or usage.other?

modules/system/rolls/wom-cast-test.js
```diff
@@ -112,31 +132,42 @@ export default class WomCastTest extends CastTest {
...
+        overcastData.usage.duration.available = overCastTable.duration[overcastData.usage.duration.count] && overcastData.available >= overCastTable.duration[overcastData.usage.duration.count].cost
       }
       if (overcastData.usage.damage) {
-        overcastData.usage.damage.available = overcastData.available >= overCastTable.damage[overcastData.usage.damage.count].cost
+        overcastData.usage.damage.available = overCastTable.damage[overcastData.usage.damage.count] && overcastData.available >= overCastTable.damage[overcastData.usage.damage.count].cost
       }
 
-      this.data.result.SL = `+${overcastData.available}`
+      // Subtract SL by the amount spent on overcasts
+      this.data.result.SL = `+${overcastData.originalSL - (overcastData.total - overcastData.available)}`
```
silentmark, line 170: i was not changing the original SL value, because i didnt see that its importent - however maybe it is, if for example this value is used in spell effects?

silentmark:
i've added two comments.

modules/system/rolls/wom-cast-test.js
moo-man: Oops, yeah, I wrote extra at one point and then changed it except here

modules/system/rolls/wom-cast-test.js
moo-man: Well, I saw two issues

- If you cast a CN 5 spell with 15 SL, you have 10 Overcasts
- If you spent 1 to +1 Damage, you now have 9
- Now, on the card, you went from 15 SL to 9 SL (should be 14)
- Then upon resetting, you went to 10 SL (should be 15)

As you say I don't think it's that important in any mechanics, but it's correct, and better to be safe

silentmark: Ok,. Now I see my mistake. I have been mostly checking this will full channeling so with CN = 0, and in this scenario issue doesn't appear

PR #1330, opened 2022-10-25
improved prefill modifiers, engagement status for range attacks

silentmark:
Improved prefill modifiers info will move the tooltip with details about prefill modifiers into dededicated location. It will also add number indicator about modifiers that actually affected the values. The same indicator will be also visible in the chat for GM users. 

In addition, a new effect - combat engagement was added, to control prefill modifiers for range attacks - when shooter is engaged - he doesn't have bonuses from range. When shooter is shooting at engaged actor - he gets -20 penalty and risk to hit another engaged actor instead. This is fully optional, as engaged condition is not applied automatically.


static/templates/dialog/dialog-constant.html
```diff
@@ -20,6 +20,15 @@
       <label title="{{localize 'DIALOG.SuccessBonusTT'}}">{{localize "DIALOG.SuccessBonus"}}</label>
       <input type="text" name="successBonus" value="{{successBonus}}" />
     </div>
+
+    {{#if prefillTooltipCount}}
+    <div class="form-group prefill-tooltip">
+      <label>{{localize "DIALOG.Details"}}:</label>
+      <div>
+        <div title="{{prefillTooltip}}">{{prefillTooltipCount}}</div>
```
moo-man, line 28: Let's use Foundry's `data-tooltip`

static/lang/en.json
```diff
@@ -903,6 +905,7 @@
     "WFRP4E.ConditionName.Unconscious" : "Unconscious",
     "WFRP4E.ConditionName.Grappling" : "Grappling",
     "WFRP4E.ConditionName.Fear" : "Fear",
+    "WFRP4E.ConditionName.Engagement": "Engagement",
```
moo-man, line 908: Let's use the book's term, **Engaged**

modules/system/rolls/test-wfrp4e.js
```diff
@@ -256,6 +258,13 @@ export default class TestWFRP {
         }
       }
 
+      if (this.options.engagedModifier) {
+        let unmodifiedTarget = target - this.options.engagedModifier;
+        if (this.result.roll > unmodifiedTarget) {
+          this.result.other.push(game.i18n.localize("ROLL.HitAnotherEngagedTarget"))
+        }
+      }
```
moo-man, line 266: Shouldn't this be in the `WeaponTest` class?

modules/actor/actor-wfrp4e.js
```diff
@@ -2586,10 +2587,10 @@ export default class ActorWfrp4e extends Actor {
     // If offhand and should apply offhand penalty (should apply offhand penalty = not parry, not defensive, and not twohanded)
     if (item.type == "weapon" && item.offhand.value && !item.twohanded.value && !(item.weaponGroup.value == "parry" && item.properties.qualities.defensive)) {
       modifier = -20
-      tooltip.push(game.i18n.localize("SHEET.Offhand"))
+      tooltip.push(`${game.i18n.localize("SHEET.Offhand")} (-20)`);
       modifier += Math.min(20, this.flags.ambi * 10)
       if (this.flags.ambi)
-        tooltip.push(game.i18n.localize("NAME.Ambi"))
+        tooltip.push(`${game.i18n.localize("NAME.Ambi")} (+${Math.min(20, this.flags.ambi * 10)})`);
```
moo-man, line 2593: I've been fine with it elsewhere, but let's have a variable here, the `Math.min(20, this.flags.ambi * 10)` being duplicated bothers me 😅

modules/system/rolls/test-wfrp4e.js
silentmark: i kept it here, because of the sizemodifier being used same place and also to handle ranged trait.

moo-man:
I've made some changes, feedback welcome!

PR #1336, opened 2022-10-29
Localizations

Txus5012:
This will allow to translate the "Target" in the scatter table. Also allows to translate the Location labels in the browser.


moo-man:
Removed the config object as I don't think it's needed (also there was a typo)

PR #1339, opened 2022-11-04
bunch of fixes (overcast, hidden opposed tests, opposed test NaN for unopposed)

silentmark:
Hi, 

bunch of fixes. 
1. whenever an overcast is used for duration - it is not applied for active effect, because test is recreated from data inside message. 
2. since mask flag is not used anymore - i fallback into hidden tokens, so now if hidden token will start opposed test - his name/icon will not be revealed. 
3. overcast for player being applicable on failed test
4. when opposed test was changed to unopposed, sometimes NaN value was showing up for me.
5. active effects now will be run in async or normal mode depending on the script.
6. there is still some issues with 'other' information in tests.

### DIFF
--- a/modules/actor/actor-wfrp4e.js
+++ b/modules/actor/actor-wfrp4e.js
@@ -1348,7 +1348,7 @@ export default class ActorWfrp4e extends Actor {
       cardOptions.speaker.scene = canvas.scene.id
       cardOptions.flags.img = this.token.texture.src; // Use the token image instead of the actor image
 
-      if (this.token.getFlag("wfrp4e", "mask")) {
+      if (this.token.document.hidden) {
         cardOptions.speaker.alias = "???"
         cardOptions.flags.img = "systems/wfrp4e/tokens/unknown.png"
       }
@@ -1357,15 +1357,15 @@ export default class ActorWfrp4e extends Actor {
     {
       let speaker = ChatMessage.getSpeaker()
       if (speaker.actor == this.id) {
+        let token = speaker.token ? canvas.tokens.get(speaker.token) : null;
         cardOptions.speaker.alias = speaker.alias
         cardOptions.speaker.token = speaker.token
         cardOptions.speaker.scene = speaker.scene
         cardOptions.flags.img = speaker.token ? canvas.tokens.get(speaker.token)?.document.texture.src : cardOptions.flags.img
-      }
-
-      if (getProperty(this.prototypeToken, "flags.wfrp4e.mask")) {
-        cardOptions.speaker.alias = "???"
-        cardOptions.flags.img = "systems/wfrp4e/tokens/unknown.png"
+        if (token?.document?.hidden) {
+          cardOptions.speaker.alias = "???"
+          cardOptions.flags.img = "systems/wfrp4e/tokens/unknown.png"
+        }
       }
     }
 
@@ -1712,7 +1712,7 @@ export default class ActorWfrp4e extends Actor {
     // if (damageType !=  game.wfrp4e.config.DAMAGE_TYPE.IGNORE_ALL)
     //   updateMsg += " ("
 
-    let weaponProperties = opposedTest.attackerTest.weapon.properties
+    let weaponProperties = opposedTest.attackerTest.weapon?.properties
     // If weapon is undamaging
     let undamaging = false;
     // If weapon has Hack
@@ -2901,20 +2901,23 @@ export default class ActorWfrp4e extends Actor {
           successBonus: args?.prefillModifiers?.successBonus,
           difficulty: args?.prefillModifiers?.difficulty
         };
-        if (!options.async)
+        if (e.script?.indexOf("await") == -1) {
           func = new Function("args", e.script).bind({ actor: this, effect: e, item: e.item })
-        else if (options.async) {
+          WFRP_Utility.log(`${this.name} > Running ${e.label}`)
+        } else if (e.script?.indexOf("await") != -1) {
           let asyncFunction = Object.getPrototypeOf(async function () { }).constructor
           func = new asyncFunction("args", e.script).bind({ actor: this, effect: e, item: e.item })
+          WFRP_Utility.log(`${this.name} > Running Async ${e.label}`)
         }
-        WFRP_Utility.log(`${this.name} > Running ${e.label}`)
-        func(args);
+        if(func) {
+          func(args);
+        }
 
         if(trigger == "targetPrefillDialog" || trigger == "prefillDialog") {
           this._handleTooltipDiff(e, preArgs, args)
           
           // If tooltip has changed, the effect modified the args, only return these effects
-          if(e.tooltip != e.label)
+          if (e.tooltip != e.label)
             appliedEffects.push(e);
         }
         else {
--- a/modules/actor/actor-wfrp4e.js
+++ b/modules/actor/actor-wfrp4e.js
@@ -3344,7 +3344,7 @@ export default class ActorWfrp4e extends Actor {
       displayString = displayString.concat(`<br><b>${game.i18n.localize("Completed")}</b>`)
     }
 
-    test.result.other.push(displayString)
+    test.preData.other.push(displayString)
 
     if (item)
       this.updateEmbeddedDocuments("Item", [item]);
--- a/modules/actor/actor-wfrp4e.js
+++ b/modules/actor/actor-wfrp4e.js
@@ -2875,7 +2875,7 @@ export default class ActorWfrp4e extends Actor {
     // WFRP_Utility.log(`${this.name} > Effect Trigger ${trigger}`)
     let effects = this.actorEffects.filter(e => e.trigger == trigger && e.script && !e.disabled)
 
-    if (options.item)
+    if (options.item && options.item.effects)
       effects = effects.concat(options.item.effects.filter(e => e.application == "item" && e.trigger == trigger))
 
 
@@ -2905,7 +2905,7 @@ export default class ActorWfrp4e extends Actor {
         difficulty: args?.prefillModifiers?.difficulty
       };
       
-      game.wfrp4e.utility.runSingleEffect(e, this, e.item, args);
+      game.wfrp4e.utility.runSingleEffect(e, this, e.item, args, options);
 
       if(trigger == "targetPrefillDialog" || trigger == "prefillDialog") {
         this._handleTooltipDiff(e, preArgs, args)
@@ -3598,8 +3598,6 @@ export default class ActorWfrp4e extends Actor {
 
     effect.origin = this.uuid;
     let duration
-
-    let multiplier = 1
     if (test && test.result.overcast && test.result.overcast.usage.duration) {
       duration = test.result.overcast.usage.duration.current;
     } else if(item.Duration) {
@@ -3607,7 +3605,7 @@ export default class ActorWfrp4e extends Actor {
     }
 
     if (duration && item.duration.value.toLowerCase().includes(game.i18n.localize("Minutes")))
-      effect.duration.seconds = duration * 60 * multiplier
+      effect.duration.seconds = duration * 60
 
     else if (duration && item.duration.value.toLowerCase().includes(game.i18n.localize("Hours")))
       effect.duration.seconds = duration * 60 * 60 
--- a/modules/actor/actor-wfrp4e.js
+++ b/modules/actor/actor-wfrp4e.js
@@ -232,6 +232,7 @@ export default class ActorWfrp4e extends Actor {
 
   }
 
+
   /** @override
    * Replaces foundry's effects getter which returns everything, to only return effects that should actually affect the actor. 
    * For example, effects from a spell shouldn't be affecting the actor who own the spell. Diseases that are still incubating shouldn't have their effects be active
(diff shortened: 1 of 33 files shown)


modules/actor/actor-wfrp4e.js
```diff
@@ -3606,19 +3584,23 @@ export default class ActorWfrp4e extends Actor {
...
+    if (duration && item.duration.value.toLowerCase().includes(game.i18n.localize("Minutes")))
+      effect.duration.seconds = duration * 60 * multiplier
 
-    else if (item.duration && item.duration.value.toLowerCase().includes(game.i18n.localize("hours")))
-      effect.duration.seconds = parseInt(item.Duration) * 60 * 60 * multiplier
+    else if (duration && item.duration.value.toLowerCase().includes(game.i18n.localize("Hours")))
+      effect.duration.seconds = duration * 60 * 60 
 
-    else if (item.duration && item.duration.value.toLowerCase().includes(game.i18n.localize("rounds")))
-      effect.duration.rounds = parseInt(item.Duration) * multiplier
```
moo-man, line 3621: > whenever an overcast is used for duration - it is not applied for active effect, because test is recreated from data inside message.

I can't seem to reproduce this. The test (and message) data should be updated with every overcast and calculate the correct value in `usage.duration`

Additionally your fix uses the same object, so why would that be different? (`duration.count` vs `duration.current`

modules/hooks/combat.js
```diff
@@ -7,12 +7,11 @@ export default function() {
   Hooks.on("deleteCombat", CombatHelpers.endCombat)
 
   Hooks.on("preCreateCombatant", (combatant, data) => {
-    let mask = canvas.tokens.get(data.tokenId).document.getFlag("wfrp4e", "mask")
+    let mask = canvas.tokens.get(data.tokenId).document.hidden;
```
moo-man, line 10: This should use `combatant.hidden`

modules/hooks/handlebars.js
```diff
@@ -27,11 +27,13 @@ export default function () {
...
 
         Handlebars.registerHelper("tokenImg", function(actor) {
-            return actor.token ? actor.token.texture.src : actor.prototypeToken.texture.src
+            let token = game.canvas.tokens.getDocuments().filter(x=>x.actorId == actor.id)[0] ?? actor.prototypeToken;
+            return token.hidden ? "systems/wfrp4e/tokens/unknown.png" : token.texture.src;
         })
 
         Handlebars.registerHelper("tokenName", function(actor) {
-            return actor.token ? actor.token.name : actor.prototypeToken.name;
+            let token = game.canvas.tokens.getDocuments().filter(x=>x.actorId == actor.id)[0] ?? actor.prototypeToken;
```
moo-man, line 35: This is really gross but I think to make it cleaner you could just use `actor.getActiveTokens()`

modules/system/combat.js
```diff
@@ -56,8 +56,12 @@ export default class CombatHelpers {
             if (turn.actor.hasSystemEffect("dualwielder"))
                 turn.actor.removeSystemEffect("dualwielder")
 
-            if (game.settings.get("wfrp4e", "statusOnTurnStart"))
-                turn.actor.displayStatus(combat.round, turn.name);
+            if (game.settings.get("wfrp4e", "statusOnTurnStart")) {
+                let token = game.canvas.tokens.get(combat.current.tokenId);
+                let nameOverride = token.document.hidden ? "???" : turn.name;
+
+                turn.actor.displayStatus(combat.round, nameOverride);
```
moo-man, line 63: Can use `game.combat.combatant` here

modules/system/opposed-test.js
```diff
@@ -161,8 +161,8 @@ export default class OpposedTest {
 
       opposeResult.other = opposeResult.other.concat(opposeResult.modifiers.message);
 
-      let attackerSL = parseInt(attackerTest.result.SL);
-      let defenderSL = parseInt(defenderTest.result.SL);
+      let attackerSL = parseInt(attackerTest.result.SL ?? 0);
```
moo-man, line 164: > when opposed test was changed to unopposed, sometimes NaN value was showing up for me.

Was this the fix? What do you mean by opposed to unopposed? Are there reproduction steps?

modules/system/opposed-wfrp4e.js
```diff
@@ -115,9 +115,14 @@ export default class OpposedWFRP {
 
   renderOpposedStart() {
     return new Promise(async resolve => {
-      let attacker = WFRP_Utility.getToken(this.attackerTest.context.speaker) || this.attacker.prototypeToken;
+      let attacker = game.canvas.tokens.get(this.attackerTest.context.cardOptions.speaker.token)?.document ?? this.attacker.prototypeToken;
       let defender
```
moo-man, line 119: why wasn't `getToken` working here?

modules/system/rolls/test-wfrp4e.js
```diff
@@ -535,7 +535,13 @@ export default class TestWFRP {
...
 
     //this.result.other = this.result.other.join("<br>")
-    this.result.other = this.preData.other.join("<br>")
+
+    if (this.preData.other.length > 0) {
+      this.result.other = this.preData.other.join("<br>");
+    }
+    else if(this.result.other && this.result.other.join) {
+      this.result.other = this.result.other.join("<br>");
+    }
```
moo-man, line 544: Not sure what these two cases are for? Couldn't it just be
```js
if (this.result.other instanceof Array)
{
   this.result.other = this.result.other.join("<br>")
}
```

modules/system/utility-wfrp4e.js
```diff
@@ -1080,9 +1080,30 @@ export default class WFRP_Utility {
...
+      }
+      if(func) {
+        func(scriptArgs);
+      }
+    }
+    catch (ex) {
+      ui.notifications.error(game.i18n.format("ERROR.EFFECT", { effect: effect.label }))
+      console.error("Error when running effect " + effect.label + " - If this effect comes from an official module, try replacing the actor/item from the one in the compendium. If it still throws this error, please use the Bug Reporter and paste the details below, as well as selecting which module and 'Effect Report' as the label.")
+      console.error(`REPORT\n-------------------\nEFFECT:\t${effect.label}\nACTOR:\t${actor.name} - ${actor.id}\nERROR:\t${ex}`)
+    }
```
moo-man, line 1106: Why was this needed exactly? There are specific triggers that can't support asynchronous execution and specific ones that can. The specific ones were given async functions already.

modules/actor/actor-wfrp4e.js
silentmark: so, the change here came during my testing and was purely cosmetical. The only potential benefit of this was to use exact duration in case there are any other homebrew in use. Anyway, here is example where problem lies: 

image

The roll is recreated from message flag after overcast click, but the flags are not updated, thats why i have added the:

      this.updateMessageFlags();

into all overcast and overcastreset messages.

modules/system/opposed-test.js
silentmark: Ok, so now i have problems to reproduce this as well, but the issue which i've been facing was related to result.SL being undefined and parseInt was returning me NaN.

modules/system/opposed-wfrp4e.js
silentmark: image


Scene was null for speaker so it was always falling back to token prototype

modules/system/rolls/test-wfrp4e.js
silentmark: Sorry for that, it was quick and dirty during session, so i forgot to fix it before commit.

modules/system/utility-wfrp4e.js
silentmark: first of all, to unify the logging errors and how effects are executed. Second of all, to allow async/await in all effects, because its not documented anywhere why and which effects may be async and which not. And even if - i dont get why only certain effects can be async? you are not awaiting any method in runEffects so it doesnt matter. In fact, you might want to look into my experimental branch https://github.com/silentmark/WFRP4e-FoundryVTT/tree/features/experimental where im trying to await all functiosn that are async, including effects.

moo-man: Any effect that gets called during actor preparation for example. Or really any time the calling function isn't async, calling effects that are won't really work / will be unpredictable.

moo-man: While I quite like unifying the code here, something about searching for "await" rubs me the wrong way, can't we still just use the `async` flag?

moo-man: If you look at the current implementation, `async` is just an option passed
```js
        if (!options.async)
          func = new Function("args", e.script).bind({ actor: this, effect: e, item: e.item })
        else if (options.async) {
          let asyncFunction = Object.getPrototypeOf(async function () { }).constructor
          func = new asyncFunction("args", e.script).bind({ actor: this, effect: e, item: e.item })
        }
```

It is underutilized however, I generally add it to a trigger when I need it to work for a specific effect and if the trigger supports it. Using callbacks was from before I added async support to effect scripts at all and tend to not do that anymore, only when I'm copy-pasting or it requires it.

`turn.actor.runEffects("endRound", combat, {async: true})`

As for what supports them, what absolutely cannot are preparation ones, such as `(pre)prepareData`, `(pre)prepareItem(s)`, `preWoundCalc`, etc. A comprehensive list would be a good thing to add to a hypothetical effects tutorial

As for why, an example would be an async `preWoundCalc` changes `sb`, `tb`, or any other of its arguments, won't happen in time for the calculation.

What could possibly support them would be `prefillDialog`? The calling function isn't async, so I think it would be unpredictable, and should probably change it to `async` first.

Point is, the `async` flag should be used and only on the ones that can fully support it. 

I would ask to hold off changing existing triggers (like `prefillDialog`) to support async for a separate PR though

modules/system/rolls/test-wfrp4e.js
moo-man: Now it still has these two cases (if/else)

```js
    if (this.preData.other.length > 0) {
      this.result.other = this.preData.other.join("<br>");
    } 
    else if (this.result.other instanceof Array) {
       this.result.other = this.result.other.join("<br>")
    }
```

But as my previous comment suggests, can't it be one if statement testing if it's an array?

modules/actor/actor-wfrp4e.js
moo-man: Do you know of a case where a homebrew duration would fail?

Additionally
```js
 if (duration && item.duration.value.toLowerCase().includes(game.i18n.localize("Minutes")))
 ``` 
Why were the units of time capitalized? I'm not sure how it could be working because lowercase duration can never match "Minutes"

Lastly, 
If the `multiplier` variable is not being used it should be removed.

File: modules/system/utility-wfrp4e.js (hunk 8 above)
Comment (silentmark) [reply modules/system/utility-wfrp4e.js:1118]: sure, i will change it, but i will leave the unified function. Once i will have more time to test my experimental branch, i will create a PR that should allow to use async functions in more broad way, and also most of them should be properly awaited. I think that this lack of awaiting asynchronous functions sometimes leads to those anoying errors with CreateEmbeddedDocument or DeleteEmbeddedDocument.

File: modules/actor/actor-wfrp4e.js (hunk 1 above)
Comment (silentmark) [reply modules/actor/actor-wfrp4e.js]: Removed, also thats a funny thing - in en.json and other translations, Minutes, Rounds etc are all translated to lower case. IT was a small bummer that i found as well.

File: modules/system/chat-wfrp4e.js [hunk 9]
```diff
@@ -414,7 +418,7 @@ export default class ChatWFRP {
     let message = game.messages.get(msgId)
     let conditionResult;
 
-    if (combatant.actor.isOwner)
+    if (combatant.actor.isOwner && game.user.isGM)
```
moo-man, line 421: Why was this added? Players applying their own conditions is preferable

modules/system/rolls/test-wfrp4e.js
moo-man: Bumping this as I'd like to merge soon

modules/system/chat-wfrp4e.js
silentmark: crap, sorry, thanks for spotting this, it was for my personal use.

modules/system/rolls/test-wfrp4e.js
silentmark: i would have to try recreate the issue one more time, but i noticed again, that in some cases, the preData.other is empty, and result.other is already an Array and not string. thats why two statements. it is something that "works for me"

moo-man: Well, as it is, it will throw an error if `this.preData.other` is a string, because `join` is not a string function. 

If we need two cases though, I think it's simpler and cleaner to combine them using `&&` conditional

silentmark: Ok, i've made another commit with proposal to get rid of preData.other as it was introduced purely for unofficialgirmoire. I hope this could solve the problem in clean way.

moo-man: Well this PR is already way out of scope for just being fixes, can you remove all changes related to `other` and put that into another PR?

silentmark: ok, changes are reverted and separate PR has been created for that

PR #1340, opened 2022-11-15
testData.other field change proposal

silentmark:


moo-man:
So while I'm not opposed to this change, I would really like to find the root cause. This shouldn't be an issue as it's a rather simple process. `preData.other` is an array and is turned into a string and placed in `preData.result` when it goes through the test evaluation process. 

It wasn't always the case though, from before when the test evaluation process was a huge mess and didn't use classes. I'm pretty sure something from that era lingers and is giving us this difficulty.

silentmark:
Sure, I will play around and try to reproduce this. I'm almost sure that problem is in some corner cases during extended test

silentmark:
Ok, i found one of the problems with "other" and extended test and it goes to the topic that i mentioned earlier - async/await inconsistency. There is following line in `handleExtendedTest`. This method is awaited in `async postTest` from main Test class.

```javascript
if (item.system.SL.current >= item.system.SL.target) {

      if (getProperty(item, "flags.wfrp4e.reloading")) {
        let actor
        if (getProperty(item, "flags.wfrp4e.vehicle"))
          actor = WFRP_Utility.getSpeaker(getProperty(item, "flags.wfrp4e.vehicle"))

        actor = actor ? actor : this
        let weapon = actor.items.get(getProperty(item, "flags.wfrp4e.reloading"))
        weapon.update({ "flags.wfrp4e.-=reloading": null, "system.loaded.amt": weapon.loaded.max, "system.loaded.value": true })
      }

      if (item.system.completion.value == "reset")
        item.system.SL.current = 0;
      else if (item.system.completion.value == "remove") {
        await this.deleteEmbeddedDocuments("Item", [item._id])
        this.deleteEffectsFromItem(item._id)
        item = undefined
      }
      displayString = displayString.concat(`<br><b>${game.i18n.localize("Completed")}</b>`)
```

but the `postTest` is not awaited in `async roll`. Also, all other `postTest` methods, when overriden, like in WeaponTest, are no longer async and this leads to errors, as the displayString is added to other collection after the card is rendered. Probably the easiest solution for this case would be to change above lines 

```javascript
        await this.deleteEmbeddedDocuments("Item", [item._id])
        this.deleteEffectsFromItem(item._id)
```

to 

```javascript
        const that = this;
        const asyncItem = item;
        this.deleteEmbeddedDocuments("Item", [asyncItem.id]).then(() => {  that.deleteEffectsFromItem(asyncItem.id)} ;
```

```javascript

  async roll() {
    this.runPreEffects();

    this.reset();
    if (!this.preData.item)
      throw new Error(game.i18n.localize("ERROR.Property"))
    if (!this.context.speaker)
      throw new Error(game.i18n.localize("ERROR.Speaker"))

    await this.rollDices();
    await this.computeResult();

    this.runPostEffects();
    this.postTest();

```

moo-man:
Could we just await postTest (and add async to the subclasses)? Seems like that would avoid this and any other issues (of which there could be several) regarding it.

silentmark:
yes, i didnt want to propose that though, because it might lead to some other race conditions elsewhere, and at this moment, i have no clue where. Personally i have a bit more customizations and one of the newest is related to streamlining async/await for almost all logic.

PR #1359, opened 2022-11-20
Hotfix

silentmark:


moo-man:
I don't like that the opposed class needs such specific WoM code, and adding more really isn't preferable. It's worth a refactor to allow attacking tests to specify whether to use opposed SL when calculating damage (there's quite a few items out there that specify 0 damage but some special effect, which isn't currently possible), but that's out of the scope of a hotfix. 

For now I've implemented an alternate fix: https://github.com/moo-man/WFRP4e-FoundryVTT/commit/0ce01ca0759c9ea53450e732cc095808080834db

PR #1421, opened 2023-01-28
sync/async effects, WoM fix, hidden token fix

silentmark:
Hi, 

this is my second attempt to make async/sync effects fully awaitable and distinguish which ones can be async. It also adds extra awaits, mostly insides of opposed test class. If you would approve this, i could move forward with few other features that i find interesting and using them in my games: 
- awaitable sockets - which allows me to request a test from player as gm and wait with rest of effect execution once the test was completed. 
- automatic response for opposed tests from minions. When i have plenty of minor opponents like goblins, ratmans or whatever, i dont like to click too much to 'roll opposed test with weapon or skill, apply damage or effects when failed'. Instead, i have small extension that allows me to mark specific tokens as 'auto-resolve' but it requires more awaitable effects and opposed tests. 

in addition: 
- there is a fix for WoM overcast template (missing closing div). 
- fix channel until ready (again). 
- fix for hidden tokens in combat - now it syncs token status with combatant status and chat cards.

### DIFF
--- a/modules/actor/sheet/actor-sheet.js
+++ b/modules/actor/sheet/actor-sheet.js
@@ -534,10 +534,10 @@ export default class ActorSheetWfrp4e extends ActorSheet {
                    do {
                     let testObject = duplicate(test);
                     let testDuplicate = test.constructor.recreate(testObject.data);
-                    if (testDuplicate.result.minormis || testDuplicate.result.majormis || testDuplicate.result.catastrophicmis) {
+                    if (test.item.cn.SL >= test.item.cn.value) {
                       break;
                     }
-                    if (test.item.cn.SL >= test.item.cn.value) {
+                    if (testDuplicate.result.minormis || testDuplicate.result.majormis || testDuplicate.result.catastrophicmis) {
                       break;
                     }
                     await testDuplicate.roll();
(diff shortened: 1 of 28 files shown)


modules/actor/sheet/actor-sheet.js
```diff
@@ -1399,25 +1402,26 @@ export default class ActorSheetWfrp4e extends ActorSheet {
...
+      else 
+        await this.actor.addCondition(condKey)
       return
     }
     if (ev.button == 0)
-      this.actor.addCondition(condKey)
+      await this.actor.addCondition(condKey)
     else if (ev.button == 2)
-      this.actor.removeCondition(condKey)
+      await this.actor.removeCondition(condKey)
```
moo-man, line 1424: These awaits seem entirely unnecessary, why await these when there's nothing coming afterwards?

modules/actor/sheet/actor-sheet.js
```diff
@@ -2318,10 +2327,12 @@ export default class ActorSheetWfrp4e extends ActorSheet {
...
 
         let weapon = item
-        if (weapon)
-          this.actor.setupWeapon(weapon, { modify: { modifier } }).then(setupData => {
+        if (weapon) {
+          let options = {modify: { modifier } };
+          this.actor.setupWeapon(weapon, options).then(setupData => {
             this.actor.weaponTest(setupData)
           });
+        }
```
moo-man, line 2335: Why separate options out when it isn't used again?

moo-man:
Instead of commenting on everything I'm just going to comment here saying I'm a bit leery about this. This seems to have just slapped `await` on... almost everything, and in many cases I don't see the reason. 

I'm going to close this, but I don't want to come off as unreceptive to the changes. So open separate PRs for your fixes. Have a another PR for your effect changes, but lets keep it focused on **effects**, or whatever is relevant there.

PR #1449, opened 2023-02-23
Feature/effects

silentmark:
Few fixes separated as per comment on https://github.com/moo-man/WFRP4e-FoundryVTT/pull/1421

there is a fix for WoM overcast template (missing closing div).
fix channel until ready (again).
fix for hidden tokens in combat - now it syncs token status with combatant status and chat cards.


modules/actor/actor-wfrp4e.js
```diff
@@ -1368,7 +1368,7 @@ export default class ActorWfrp4e extends Actor {
       cardOptions.speaker.scene = canvas.scene.id
       cardOptions.flags.img = this.token.texture.src; // Use the token image instead of the actor image
 
-      if (this.token.hidden) {
+      if (this.token.document?.hidden ?? this.token.hidden) {
         cardOptions.speaker.alias = "???"
         cardOptions.flags.img = "systems/wfrp4e/tokens/unknown.png"
       }
```
moo-man, line 1374: I removed this in https://github.com/moo-man/WFRP4e-FoundryVTT/commit/7cd770ad8f435ed2d8e18cc7a97ff5e721fb5b57

`this.token` refers to a `TokenDocument` instance already, what's the purpose of checking for `document`. It's always undefined. 
image

modules/hooks/combat.js
```diff
@@ -6,22 +6,41 @@ export default function() {
...
+      else if (combatant && !mask && !combatant.hidden && combatant.name == "???") {
+        let data = {};
+        data.img = token.texture.src;
+        data.name = token.name;
+        await combatant.update(data);
+      }
     }
-  })
+  });
+
```
moo-man, line 43: As per the previous PR, these don't really need to be async, would prefer to just use one update at the end

modules/item/item-wfrp4e.js
```diff
@@ -1583,8 +1583,10 @@ export default class ItemWfrp4e extends Item {
     let skill
     if (this.type == "weapon") {
       skill = skills.find(x => x.name.toLowerCase() == this.skill.value.toLowerCase())
-      if (!skill)
-        skill = skills.find(x => x.name.toLowerCase().includes(`(${this.WeaponGroup.toLowerCase()})`))
+      if (!skill) {
+        let weaponGroupName = game.wfrp4e.config.weaponGroups[this.weaponGroup.value];
+        skill = skills.find(x => x.name.toLowerCase().includes(`(${weaponGroupName.toLowerCase()})`))
+      }
```
moo-man, line 1589: What's the purpose here? `this.WeaponGroup` is the same as `game.wfrp4e.config.weaponGroups[this.weaponGroup.value];`

silentmark: you've got me here. Probably i was debugging something and didnt notice the lowercase field vs uppercase property getter. I will revert this one.

modules/actor/actor-wfrp4e.js
silentmark: I'm pretty sure, that i was getting some invalid data here - that this.token.hidden sometimes was not present, but maybe that's because im not so fluent with Foundry Api. Another option it is related to Token Action hud behavior. I will check that and if reproducable - add another comment or remove the change.

moo-man: Any update on this? Would like to merge soon

silentmark: i removed that, couldnt reproduce anymore on clean system, must have been some other changes or modules i had.

modules/hooks/item.js
```diff
@@ -53,18 +53,17 @@ export default function () {
...
-            
-            item.actor.update({ "system.status.wounds.value": newWounds });
-            ui.notifications.notify(`${item.wounds.value} ${game.i18n.localize("CHAT.CriticalWoundsApplied")} ${item.actor.name}`)
+          }
+        } else if (item.wounds.value.toLowerCase() == "death") {
+          newWounds = 0;
         }
-
+        item.actor.update({ "system.status.wounds.value": newWounds });
+        ui.notifications.notify(`${item.wounds.value} ${game.i18n.localize("CHAT.CriticalWoundsApplied")} ${item.actor.name}`)
```
moo-man, line 66: What is the purpose of this change? It seems like a partial reversion of https://github.com/moo-man/WFRP4e-FoundryVTT/commit/6934611265d87ce8103430085f1ab4c18b52374c

I don't see any mention of what this fixes in the previous PR either

modules/system/config-wfrp4e.js
```diff
@@ -1697,13 +1701,11 @@ WFRP4E.conditionScripts = {
         let effect = actor.hasCondition("ablaze")
         let value = effect.conditionValue;
 
-        let leastProtectedLoc;
         let leastProtectedValue = 999;
         for (let loc in actor.status.armour)
         {
             if (actor.status.armour[loc].value != undefined && actor.status.armour[loc].value < leastProtectedValue)
             {
-                leastProtectedLoc = loc;
```
moo-man, line 1706: Revert this change as `leastProtectedLoc` is now used, see https://github.com/moo-man/WFRP4e-FoundryVTT/commit/5781681e737255ea91199f2b4e9e3dbe5113e0a6

modules/system/rolls/prayer-test.js
```diff
@@ -126,11 +126,11 @@ export default class PrayerTest extends TestWFRP {
   
   // @@@@@@@ Overcast functions placed in root class because it is used by both spells and prayers @@@@@@@
   async _overcast(choice) {
-    if (this.result.overcast.usage[choice].AoE)
-    {
-      return ui.notifications.error(game.i18n.localize("ERROR.PrayerAoEOvercast"))
+    if (this.result.overcast.usage[choice].AoE) {
+      ui.notifications.error(game.i18n.localize("ERROR.PrayerAoEOvercast"))
+    } else {
+      await super._overcast(choice)
```
moo-man, line 132: I don't think await `super._overcast` is necessary here

silentmark: are you sure? i think that default overcast options for prayer should be applied.

modules/hooks/item.js
silentmark:
```javascript
        if (Number.isNumeric(item.wounds.value))
        {
          if (item.wounds.value.toLowerCase() == "death")
            newWounds = 0;
            newWounds = item.actor.status.wounds.value - Number(item.wounds.value)
            if (newWounds < 0) newWounds = 0;

            item.actor.update({ "system.status.wounds.value": newWounds });
            ui.notifications.notify(`${item.wounds.value} ${game.i18n.localize("CHAT.CriticalWoundsApplied")} ${item.actor.name}`)
        }
```

I never really understood this original piece of code. First of all, i was confused about isNumeric function that is not a standard JS (probably comes from jQuery or something extension?). Second of all, why compare to == "death" is inside of IsNumeric condition? besides that i dont have anything else to add.

modules/system/config-wfrp4e.js
silentmark: reversed

modules/hooks/item.js
moo-man: Yeah good point, it's based off some very early code, and I didn't realize `isNumeric` isn't native javascript

My main concern is just making sure it handles UiA's trivial critical hits (why they didn't just use 0 instead of T for the wounds value i'll never understand).

modules/system/rolls/prayer-test.js
moo-man: Oh I specifically just mean the `await` which seems to be the only real change

PR #1626, opened 2023-06-10
Item Browser Caching Solution

JustArion:


modules/apps/wfrp-browser.js
```diff
@@ -178,14 +178,30 @@ export default class BrowserWfrp4e extends Application {
...
+    BrowserWfrp4e.prototype.documentCache = BrowserWfrp4e.prototype.documentCache || {};
+  
+    async function cacheDocuments(pack, documents) {
+      BrowserWfrp4e.prototype.documentCache[pack.collection] = documents;
+    }
+    
+  
+    async function getCachedDocuments(pack) {
+      if (BrowserWfrp4e.prototype.documentCache.hasOwnProperty(pack.collection)) {
+        return BrowserWfrp4e.prototype.documentCache[pack.collection];
```
moo-man, line 190: Using prototype seems a bit awkward, I think something like `game.wfrp4e.documentCache` seems clearer?

JustArion: This can be done, I'll update the PR to include this change.

JustArion: I've gone ahead and added the suggested location for caching. The only current delay that still persists is the delay from rendering the browser, the PR doesn't address that currently.

PR #1668, opened 2023-07-04
Adding Compendium Name and Module Name to tooltip. Resolves #1639

### LINKED ISSUE #1639: [WFRP4e System] Item Browser doesn't show Compendium
When I browse weapons using the Item Browser, I can see 3 Warhammers in compendiums.

1 is Core
1 is Up in Arms
third one I have no idea.

I would like an option to be able to quickly see which item is from which Compendium.
It could be done either by small cursive text under the item's name, or in tooltip when hovering over the "compendium" icon. 

It will also hugely help with quickly finding proper versions of Talents, Spells etc. that have multiple version across multiple Modules (and can also help with own custom items)<br/>**From**: acaea3d8-d962-4579-a9bc-9771c4c5be66@anonaddy.me<br/>foundry: 11.302<br/>wfrp4e: 6.5.9<br/>wfrp4e-archives1: 3.0.2<br/>wfrp4e-archives2: 3.0.6<br/>wfrp4e-core: 3.1.4<br/>wfrp4e-rnhd: 3.0.7<br/>wfrp4e-up-in-arms: 3.0.6<br/>wfrp4e-wom: 3.0.5<br/>wfrp4e-zoo: 3.0.2<br/>Active Modules: darkraven-games-soundscapes-free, dice-so-nice, forien-armoury, forien-quest-log, forien-unidentified-items, foundryvtt-simple-calendar, lib-wrapper, michaelghelfi, polyglot, popout, tabletop-rpg-music, token-action-hud, tomcartos-ostenwold, wfrp4e-gm-toolkit

Forien:


static/templates/browser/browser.hbs
```diff
@@ -490,7 +490,7 @@
             <a class = "name-icon">
               <div class = "item-name">{{item.name}}</div>
               {{#if item.compendium}}
-              <div class="icon" title="Compendium Item" style="background-image: url(systems/wfrp4e/ui/sidebar-button-compendium.webp)"></div>
+              <div class="icon" title="Compendium Item ({{item.compendium.metadata.label}}@{{item.originModule}})" style="background-image: url(systems/wfrp4e/ui/sidebar-button-compendium.webp)"></div>
```
moo-man, line 493: Not sure I like this format. `compendium.metadata.label` should include the module, right? LIke `Items (Up in Arms)`

Note that I will be converting all the Core compendium packs into one pack per type, so the `Spells`, `Talents`, etc. compendium packs will eventually be `Items (Core Rulebook)`

Forien: > Not sure I like this format. `compendium.metadata.label` should include the module, right? LIke `Items (Up in Arms)`

That's only true for Compendiums you make. For example if I make a module with Compendium and Compendium is literally called "Items", then `compendium.metadata.label` will show `Items`. 

I made it thinking about 3rd party modules and people's own custom Compendiums as well.

I agree that format `Items@Forien's Armoury` is not perfect, but couldn't think of any better way to do it in tooltip. 


> Note that I will be converting all the Core compendium packs into one pack per type, so the `Spells`, `Talents`, etc. compendium packs will eventually be `Items (Core Rulebook)`

Please at least do folders inside :D

Forien: I guess Module is not _that_ important if we assume all fan modules will have unique names of Compeniums then it shouldn't be an issue.

moo-man: > Please at least do folders inside :D

Of course! That's the point

But that is true, but instead of `Label@Module` let's just do `Label (Module)`

Forien: Alright. Gonna go sleep now, will update the PR tomorrow

moo-man: Hm, that would do something like `Items (Forien Armoury) (Forien's Armoury)` though, so I'm not sure either.

Forien: > Hm, that would do something like `Items (Forien Armoury) (Forien's Armoury)` though, so I'm not sure either.

I'm slowly shifting towards leaving just Label. Official modules will have module Included in the label, I can follow suit with my module and we can just assume others will as well.

Unofficial Grimoire prefixes all Compendiums with "UG" as well.

Issue #1621, opened 2023-06-08
[WFRP4e System] Alternative autosuccess numbers


grz0:
**Is your feature request related to a problem? Please describe.**
Allow setting alternative autosuccess numbers. Related rule from the core rulebook:

> Not everybody likes a 5% chance of auto-success on any Test. Similarly, not everyone likes the same chance of auto-failures. If this describes you, then simply change the numbers to your taste. The most common alternative is 96–00 always fails, and 01 always succeeds.

PR #1675, opened 2023-07-09
Added autosuccess/autofailure ranges  

Forien:
Added settings that allow changing Autosuccess/Autofailure ranges as described on the page 150 of Core Rulebook.

Resolves moo-man/WFRP4e-FoundryVTT#1621


moo-man:
Parking this while I assess any effect scripts that need changing.

Issue #1623, opened 2023-06-09
[WFRP4e System] Modified Critical rolls going out of bounds


moo-man:
When a modified critical roll is made, such as when wounds go below 0, but greater than negative Toughness Bonus, then a Critical Roll, modified by -20 occurs. According to Page 172, the minimum value would be 1, however the system does generate negative values under these circumstances. <br/>**From**: marauder13#5427<br/>foundry: 10.291<br/>wfrp4e: 6.5.8<br/>wfrp4e-altdorf: 2.0.2<br/>wfrp4e-archives1: 3.0.1<br/>wfrp4e-archives2: 3.0.5<br/>wfrp4e-core: 3.1.3<br/>wfrp4e-dotr: 3.0.3<br/>wfrp4e-eis: 3.1.5<br/>wfrp4e-middenheim: 3.0.1<br/>wfrp4e-pbtt: 3.0.0<br/>wfrp4e-rnhd: 3.0.6<br/>wfrp4e-starter-set: 3.1.4<br/>wfrp4e-ua1: 3.0.1<br/>wfrp4e-ua2: 2.0.0<br/>wfrp4e-up-in-arms: 3.0.5<br/>Active Modules: about-time, combat-carousel, dice-calculator, drag-ruler, fateroll, foundryvtt-simple-calendar, reclusive-cartographer-maps, socketlib, token-action-hud, token-mold, tomcartos-ostenwold, wfrp4-wfrp4e-npc-generator, wfrp4e-character-details, wfrp4e-gm-toolkit, wfrp4e-random-treasure

PR #1676, opened 2023-07-09
Fixed critical going out of bounds on -20 modifier

Forien:
Resolves moo-man/WFRP4e-FoundryVTT#1623


moo-man:
Let's just change the default behavior to be minOne = true

Forien:
> Let's just change the default behavior to be minOne = true

Wasn't sure if it wouldn't affect something else, so wanted to be on the safe side. But yeah, that works too, gimme sec :)

PR #1679, opened 2023-07-09
Fixed skill advancement modifier not being used when calculating advancement range cost

Forien:
Fixed skill advancement modifier not being used when calculating advancement range cost 

Resolves moo-man/WFRP4e-FoundryVTT#1553


Forien:
I made second commit that changes styling on the code (mainly setting same indentation level as neighbouring functions), so it can be hard to see the diff there.

Changes to code can be viewed here: https://github.com/moo-man/WFRP4e-FoundryVTT/pull/1679/commits/4055344dc86ca3c87effb20ec52dcec021722004

PR #1689, opened 2023-07-11
Some CSS to make Item Piles more readable. Probably not perfect. Config screen could use more love.

Forien:


Forien:
Don't merge it yet, I found several places to fix. Will add another commit tomorrow

Forien:
Okay, in my eyes it's done. Could use some polish with sliders when configuring Item Piles or some other things, but most imporantly it's all readable and the stuff that Players see looks in my opinion decent.

PR #1694, opened 2023-07-12
Add worded modifiers ("Half", "Twice" etc.) for Spells

### LINKED ISSUE #1684: [WFRP4e System] "Half" Modifier not working in Spell's attributes like Range, Duration, Target/AoE
For example `Death's Release`, `Captivating Flame` or other spells in WoM have "Half" in Range.

Yet spells when "unrolled" / "uncollapsed" in the "Magic" tab on the character sheet will show something like "Half 35 yards".

Support for "Half, Quarter" and other word modifiers for Range, Duration, AoE/Target would be appreciated, especially when creating custom spells.

<br/>**From**: forien<br/>foundry: 11.304<br/>wfrp4e: 6.5.9<br/>wfrp4e-core: 3.1.4<br/>wfrp4e-up-in-arms: 3.0.6<br/>wfrp4e-wom: 3.0.5<br/>Active Modules: ATL, dfreds-chat-pins, dfreds-effects-panel, forien-armoury, forien-custom-utility, lib-wrapper, socketlib, times-up, wfrp4e-gm-toolkit, wfrp4e-unofficial-grimoire

Forien:
Resolves #1684

Caveat: I only realized that near the end, but I guess my solution will break if someone does Range like `Half Willpower Bonus + Twice Intelligence Bonus`. It will completely ignore the `Twice`... so not sure it's the best solution then.

On the other hand I don't know if there exists a spell that would consider this, so not sure if it's worth time chasing a very specific edge case...

### DIFF
--- a/modules/item/item-wfrp4e.js
+++ b/modules/item/item-wfrp4e.js
@@ -1191,7 +1191,7 @@ export default class ItemWfrp4e extends Item {
       return value;
 
     // Check for generic word modifiers such as "half", "triple" etc.
-    let modifier = WFRP_Utility.processWordedModifier(ammoValue.toLowerCase(), value);
+    let modifier = WFRP_Utility.processWordedModifier(ammoValue, value);
     if (modifier !== false)
       return modifier;
     else // If the range modification is a formula (supports +X -X /X *X)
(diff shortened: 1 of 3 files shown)


moo-man:
This should probably receive a lot of scrutiny and testing to ensure all the spells still correctly compute (and this also may affect translations).

Forien:
I tried to make sure existing translations are not impacted. 

As for spells, from what I've tested it works fine, but yeah, more testing is better. It's not urgent issue, so PR can sit here for a while, no worries

silentmark:
@Forien, sorry for jumping in, but is this really necessary? you can have any basic arythmetic operations on integers, and introducing words like twice/half will complicate life for i.e. Polish translation. English has much easier rules of declinazione, in contrary to Polish, where such wording would either sound badly or would be much harder to parse.

Forien:
> @Forien, sorry for jumping in, but is this really necessary? you can have any basic arythmetic operations on integers, and introducing words like twice/half will complicate life for i.e. Polish translation. English has much easier rules of declinazione, in contrary to Polish, where such wording would either sound badly or would be much harder to parse.

`Death's Release`, `Captivating Flame` and other spells in WoM are specifically worded `Half Willpower yards` or `Half Initiative yards` and their FVTT counterparts have Ranges worded as that in official WoM module too.

If Official Module has `Half Willpower yards` but then system itself screws stuff up and displays range as `Half 40 yards` then something is not right. And it will interfere with macros and 3rd party modules that might use Spell's range, even if they only extract numeric value, it will not be accurate true value.

Your argument could be used on ammunition as well. Why have "twice weapon" or "half weapon" if we have `*` and `/` operators?

I think if we allow worded modifiers in one place where Rulebooks use them, we should allow in all places Rulebooks use them (and for further consistency, if we allow users to use `Half` etc in Spell's Range, they will be confused if we don't allow it in Duration).


### Translation concerns
For Polish, it's not that hard to translate those words. "Czwarta", "Trzecia", "Połowa", "Dwukrotność", "Trzykrotność" etc. 

I would argue (and agree) that lack of ability to properly decline words for Characteristics is way bigger issue here. "Half Willpower Bonus" would probably need to end up like "Połowa Siła Woli metrów" which sounds just plain bad. 

**_BUT!_**, nobody stops translators to (using Babele or something) completely change "Half Willpower yards" to "Siła Woli / 2 metrów". Remember that those things are within Spell Items itself, not within main translation file. And support for arithmetics is not taken away, so basically, **nothing in this Pull Request changes for translations in comparison to how things work now**. All current features work. If you translated `as weapon` and `half weapon`, it will still register properly. If you used arithmetics in translations, this feature is still there. 

The only difference is that it now **additionally** supports worded modifiers for new things, because C7 started adding those to Spell Ranges and _that_ is currently not supported by this system

silentmark:
thanks for explanation, as long as it's backward - compatibile, i feel more secured, especially that we are close to end with PL translation. However, i still think it would be better to change the WoM module spells with ```Initiative / 2 yards``` instead.

Forien:
Accidentally deleted branch on my end, sorry :D

PR #1733, opened 2023-08-13
Send Test Requests over Socket and effects improvements. 

silentmark:
Hi, 

this is the actual improvement that i wanted to propose, and previous updates were only a prepartion steps. I would like to be able as GM, send requests for Tests to players. This way, a GM instead of saying 'Roll Perception' could easily open a test for given player and ask him to roll. Of course, its a minor improvment, however, for active effects it could open many new possibilities. There are many rules in the book saying "do this and that when test fails" and one example could be Cool Test on Blackpowder Weapon shoot:

```
if (args.test.weapon?.id == this.item.id) {
    if (args.test.data.context.targets && args.test.data.context.targets.length > 0) {
        let targets = args.test.data.context.targets;
        for(let i = 0; i < targets.length; i++) {
            let tokenId = targets[i];
            let token = game.canvas.tokens.get(tokenId.token); 
            if (token) {
                let test = await token.actor.setupSkill("Opanowanie", { title: "Test Opanowania - Wystrzał z broni palnej",  absolute : {difficulty : "average"}});
                if (!test.result.outcome) {
                    await test.roll();
                }
                if(test.result.outcome == "failure") { 
                    const html = `<div>${token.actor.name} Recieve @Condition[Broken] because of gunshoot.</div><div class="apply-conditions"><a class="chat-button apply-condition" data-cond="broken">Dodaj Stan Panika</a></div>`
ChatMessage.create({ content: html });
                }
            }
        }
    }
}
```

With this option enabled, this effect running on RollWeaponTest would open a modal dialog for character's owner and on fail - show message on chat to apply Broken, or even - automatically apply Broken. 

There is also another change in philosophy here - the effect is always run in context of GM instead of a user. It has one disadvantage in context of few existing effects - whenever an effect shows a notification - it will be presented to GM instead of user.


moo-man:
Closing this due to the effect refactor, let's discuss reimplementing it with the new system.

PR #1837, opened 2023-12-05
Add clickable property link for rich text editors

Forien:
Allows using `@Property[key]{label)` format to create a clickable links.

When clicked, they will push Property's description to chat, similarly to `@Condition` syntax. 

image


Forien:
I didn't think of this as a replacement of current "linking to journal", more like just creating a support for this syntax in case some GMs/players would like to use it.

PR #1841, opened 2023-12-10
Fixed `ammoRange` related errors

Forien:
This Pull Request fixes `ammoRange` typo (it has never been declared and I believe by analyzing the code it should be `ammoValue`. 

Additionally I've added graceful fallback in a form of extra `try…catch` wrapper in order to not block opening Actor Sheet, while still providing obvious Error for the users to know that something went wrong. 

Issue could be encountered when using worded modifiers that were not properly translated (for example item said `come arma` in Italian, but client was set to `English`), or when using multipliers like `*5` or `/5` as ammo range. 

Should be fully compatible with #1694 Pull Request.

### DIFF
--- a/modules/item/item-wfrp4e.js
+++ b/modules/item/item-wfrp4e.js
@@ -1207,14 +1207,19 @@ export default class ItemWfrp4e extends Item {
       value *= 2;
     else // If the range modification is a formula (supports +X -X /X *X)
     {
-      try // Works for + and -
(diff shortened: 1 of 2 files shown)


Forien:
Merged in as #1849 into [`effect-refactor`](https://github.com/moo-man/WFRP4e-FoundryVTT/tree/effect-refactor) branch

PR #1845, opened 2023-12-13
Add `Melee (Parry)` to skill list in Roll Dialog if weapon is Defensive

Forien:
## Change no.1
CRB states that 
> Any one-handed weapon with the Defensive Quality can be used with Melee (Parry)

So my first change in commits (969ce525f2adce442665a00dda7c8a22d93d0bf9, 4b7f63632aab835a7273af9a27b557af99b555c1) adds option to roll dialog, where `Melee (Parry)` will appear if character uses a weapon with `Defensive` quality **and** if that character has `Melee (Parry)` skill. 

`Melee (Parry)` will be set as default if its `total` is higher than standard skill's. 
image

## Change no.2
My other commits (5b95e74b166b38dbf40649b0fb8b2112a39a8127, along with 4f5b63133c7c7077eb5a4d0221fd9746f0321d6f) tries to mitigate an issue reported by Canis here: https://discord.com/channels/787396882875416606/787397718339223603/1184475915783249990

> So, it does seem that the system does not apply the off hand penalty when attacking with a parry weapon.

So I added `this.isOpposing` check, to only mitigate penalty when **Opposing** tests with Parry weapon.

Furthermore, to make it more compatible with first commit, I added a `usesMeleeParry` option so I can check if Melee Parry is set as default, because the offhand penalty should not apply (quoting CRB) "When using Melee (Parry)", not when weapon group is `Parry`. Otherwise, it behaves as it did.

### DIFF
--- a/modules/actor/actor-wfrp4e.js
+++ b/modules/actor/actor-wfrp4e.js
@@ -2675,10 +2675,10 @@ export default class ActorWfrp4e extends Actor {
         item.offhand.value &&
         !item.twohanded.value &&
         !(
+          this.isOpposing &&
           item.properties.qualities.defensive &&
           (item.weaponGroup.value === "parry" || !!options.usesMeleeParry)
-        ) &&
-        this.isOpposing
+        )
     ) {
       modifier -= 20;
       tooltip.push(`${game.i18n.localize("SHEET.Offhand")} (-20)`);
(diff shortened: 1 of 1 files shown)


Forien:
The only thing that is **really important** in my opinion is adding the `this.isOpposing` to the `if` in the `ActorWfrp4e.attackPrefillData()` method.  The rest is "nice to have".


On the other hand, now with changes I proposed, when someone is using offhand weapon with `Melee (Parry)` skill by default, they get no penalty for opposing enemy attacks. But changing skill back to `Melee (Basic)` does not bring the penalty back, because penalty is not calculated on the roll dialog.
And from what I can see, `tooltip` is no longer an array and there is no built-in support for rebuilding the `tooltip` after Dialog has been rendered.  

I could do a relatively small refactoring, which would allow Dialog Application to build tooltip dynamically, based on the changes on the Dialog. 
In general I think it could help in the future, even if for now it would barely be used. 

So let me know if I should draft a refactor for Tooltips etc.

moo-man:
I don't think I'm going to be releasing an update before the effect refactor. If you want this to be accepted I'd target that branch

PR #1846, opened 2023-12-13
awaitable sockets and test requests PoC

silentmark:
Hi, 

Here is refactored awaitable sockets + awaitable test requests. I have also included few minor fixes/improvements that i found during testing.


moo-man:
> I have also included few minor fixes/improvements that i found during testing.

Please filter these out into a separate PR. I've suggested this in the past, if a PR is going to be implementing an involved feature like this they need to be focused on that feature.

PR #1847, opened 2023-12-13
combat refactor

silentmark:
Some improvements in combat scripts. Better handling current combatant / previous combatant for turn change. This helps to avoid scripts execution when you move back and forth on combat tracker. I have also unified the way how effects and combat scripts are executed - always first execute scripts and then effects. This is useful as in fact, i started to add new scripts in my module to the list of combat scripts.

### DIFF
--- a/modules/actor/actor-wfrp4e.js
+++ b/modules/actor/actor-wfrp4e.js
@@ -1390,7 +1390,7 @@ export default class ActorWfrp4e extends WFRP4eDocumentMixin(Actor)
     let msg = game.i18n.format("CHAT.DiseaseFinish", { disease: disease.name })
 
     if (disease.system.symptoms.includes("lingering")) {
-      let lingering = disease.effects.find(e => e.name.includes(game.i18n.localize("WFRP4E.Symptom.Lingering")))
+      let lingering = disease.effects.find(e => e.name.includes("Lingering"))
       if (lingering) {
         let difficulty = lingering.name.substring(lingering.name.indexOf("(") + 1, lingering.name.indexOf(")")).toLowerCase()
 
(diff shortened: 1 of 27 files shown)


moo-man:
`always first execute scripts and then effects`

I'm not sure I understand this, effects don't "execute" anymore, it's always their scripts. Is there an example of what you mean here?

silentmark:
Look at ```checkDiseases``` method - it has ```await Promise.all(turn.actor.runScripts("endRound", combat, {async: true}))``` at the end - and this method is the last method of endRound scripts - so its executed at the end, but if for example you would add another script to this endRound list - it wouldn't be true anymore.

moo-man:
The only thing changed about that method in this PR is that it has become async, it does not call any scripts. Am I looking in the wrong place?

silentmark:
for this particular example, the async is made only for consistency with other methods. as for main change, the ```await Promise.all(turn.actor.runScripts("endRound", combat, {async: true}))``` has been moved to ```static async updateCombat(combat, changes, context)``` this way its more readable and in case of extension of ```CombatHelpers.scripts.endRound``` - it will remain consistent. First scripts, then effects.

moo-man:
I will also point out that much of this is code that hasn't been refactored yet. Ideally conditions will be handled no differently than any other effect that has scripts, instead of referencing a separate script via condition key and specifically calling it.

silentmark:
I hope in case of refactoring of this part, you could take into account some of my suggestions ;-)

moo-man:
It simply depends on how easy it is to move condition scripts into effect scripts. This would result in removing `checkStartTurnConditions`, `checkEndTurnConditions`, and `checkEndRoundConditions` entirely, but otherwise there probably won't be conflicts.

silentmark:
the main problem i see there is fact, that those effects would have to be present and duplicated on every actor, which may lead to number of problems i guess. From the other hand - this mechanism to inject into combat flow with custom scripts is quite useful, the main problem i had at the beginning was inconsistency on combatant change and multiple execution of effects when switching the combatant order.

moo-man:
> those effects would have to be present and duplicated on every actor

The effects are already on every actor. If three different tokens have Ablaze, they all have an independent Ablaze effect, the difference is that the script is not within the effect, it is shared between them in config.

PR #1848, opened 2023-12-13
awaitable sockets and test requests PoC

silentmark:
Refactored awaitable sockets + awaitable test requests PoC (characteristics and weapon working so far)


moo-man:
From a quick glance and kneejerk reaction, awaitable sockets seem fine, but I don't know if I see this "test request" as something I want to include in the system. It seems like module territory to me (it would seem like a good inclusion into GM Toolkit).

silentmark:
I hope that in the end you could include this feature. From one hand it can be simply disabled and from other - it can help, among other things, new players to faster jump into system itself (GM can request a roll if they forget about something, also from code), and secondly, this feature would be quite hard to impleleme as module

moo-man:
I certainly understand the utility, but I think modularity (being one of Foundry's main attributes) is something that should be embraced here. I feel like this is a great "extra feature" for GMs that want it, and they can seek it out via modules, something the userbase is quite familiar with. Granted I haven't tried to implement it as a module, but it doesn't seem that bad.

One of the first things that stands out to me is the block of code that's going to be pasted in every setup function
```js
    if (owner.id != game.user.id && isSocketTest) {
      await this._preSetupSocketTest(owner);
      let dialogClassName = CharacteristicDialog.name;      
      let payload = { dialogData, dialogClassName, userId: game.user.id, actorId: this.id };
      return game.wfrp4e.utility.setupSocket(owner, payload);
    } else {
      return this._setupTest(dialogData, CharacteristicDialog)
    }
```
This isn't something I particularly like, and I notice that the only difference between the two you've implemented is
```js
dialogData.data.weapon = weapon.toObject();
```
This really indicates to me that this should be done differently: move socket handling up sooner, to the **sheet level**, so you can send "what test to setup" via socket instead of "here's all the test data you need" which I think has multiple advantages. For one, you can remove this ugly `crtlKeyPressed` global listening, simply use `ev.crtlKey`. Secondly, as a module, you can simply override the actor sheet functions (such as `_onCharClick` or `_onSkillClick`, test for sockets, if true -> send what test to perform via socket, else if false just call `super.****` 

As a compromise, we can talk about what to keep here in the system to make a module easier to create, which is probably the specific socket handling.

silentmark:
There is bunch of reasons why I did it this way. First approach was to send the socket data inside ```_setupTest``` but since there are actual objects in data instead of simple dataObjects which can be serialised I moved it up. Secondly there are more things that will differentiate, but I only did and tested two kinds of tests. The spell/prayer tests will require different thing to pass
 Maybe I will find a way to generalize this method but not yet.
Also, the reason why I don't want to embed it on sheet is because of other modules setting up tests, i.e. token action HUD. I would also like to take into account this feature in other cases like effects or my own private module that I might publish - automatic fight which allows to automatically respond to attacks with opposed defence test or even perform attack on its own.

Jagusti:
FWIW I do have a different request workflow brewing as an extension of the Make Group Test feature of GM Toolkit. It's basic functionality is committed on the `dev` branch. 

There are some QOL improvements to make but it's functional now without changes needed to the wfrp4e system. I might just release it as-is sooner rather than later.

I'm not looking to change the workflow on it (I want too keep it consistent with other Toolkit flows), or increase the scope, which essentially handles skill tests and fallback characteristics tests, which are the main use cases. 

Other modules can probably tap into the request functionality via `game.gmtoolkit.grouptest.runActorTest` though I haven't checked and may need to expose that first. That also stashes the test results in a temporary world setting so you can do things with it if needed (I use it to post summary group test results in chat).

To support the use case of better onboarding new players, I'd probably look at tours. With some unlicensed wizardry you could probably push the roll dialog to a player using the new GM Toolkit functionality and then trigger a tour with Foundry's baked-in functionality to guide them through options, selecting relevant talents and what not.

silentmark:
Ok, so i've removed the socket tests. If the rest of PR would be approved, i would try to implement it as part of the module or open another PR just in case for the future.

modules/system/socket-handlers.js
```diff
@@ -1,97 +1,93 @@
...
 
     static register()
     {
-        game.socket.on("system.wfrp4e", data => 
+        game.socket.on("system.wfrp4e", async data => 
         {
-            this[data.type]({...data.payload}, data.userId);
-        });
-    }
+            if (data.userId != game.user.id && data.userId != "ALL") return;
```
moo-man, line 20: Shouldn't the case of `userId` being empty be the same as `ALL`? There are plenty of old socket requests in the code that don't specify `userId`, wouldn't this line break all of them? Or is there something I'm missing that handles that.

silentmark: i think i can fallback the lack of userid into ALL.

silentmark: Added fallback

PR #1850, opened 2023-12-14
bunch of improvements

silentmark:
- better ACE editor (javascript support with latest ECMA, resizable window
- fixed two places where i noticed old script execution
- fixed computeDerived where missing arguments were passed to super function (seems that those are not used at all anyway). 
- cli bump, 
- Lingering effect name translation, 
- added combat round information to templates for future auto-removal. 
- error handling and removed unused variable.


modules/actor/actor-wfrp4e.js
```diff
@@ -1790,11 +1790,14 @@ export default class ActorWfrp4e extends WFRP4eDocumentMixin(Actor)
   }
 
 
-  applyTerror(value, name = undefined) {
+  async applyTerror(value, name = undefined) {
     value = value || 1
     let terror = duplicate(game.wfrp4e.config.systemItems.terror)
     terror.flags.wfrp4e.terrorValue = value
-    return game.wfrp4e.utility.applyOneTimeEffect(terror, this)
+    let scripts = new EffectWfrp4e(terror, {parent: this}).getScripts();
```
moo-man, line 1797: This class doesn't have a `getScripts` function, like actors and items. You can simply call the `scripts` getter on an effect and it will construct all the script objects.

modules/model/item/components/base.js
```diff
@@ -132,8 +132,8 @@ let fields = foundry.data.fields;
         data.properties = [];
         data.description.value = data.description.value || "";
         data.description.value = await TextEditor.enrichHTML(data.description.value, htmlOptions);
-        data.targetEffects = this.parent.effects.filter(e => e.application == "apply")
-        data.invokeEffects = this.parent.effects.filter(e => e.trigger == "invoke")
+        data.targetEffects = this.parent.effects.filter(e => e.application == "target")
+        data.invokeEffects = this.parent.effects.filter(e => e.trigger == "manual")
```
moo-man, line 136: These can be removed as `e.application` and `e.trigger` don't exist anymore. I'll be adding the new ways of doing this soon.

modules/system/overrides.js
```diff
@@ -147,7 +147,7 @@ export default function () {
 
 
   Token.prototype.incrementCondition = async function (effect, { active, overlay = false } = {}) {
-    const existing = this.actor.actorEffects.find(e => e.conditionKey === effect.id);
+    const existing = this.actor.effects.contents.find(e => e.conditionKey === effect.id);
```
moo-man, line 150: this could probably be changed to `this.actor.hasCondition(effect.id)`

static/css/wfrp4e.css
```diff
@@ -14760,23 +14760,21 @@ textarea,
...
 
-.active-effect-sheet .ace-editor > *{
-  height: 300px;
+.script-config .ace-editor {
+  height: 100% !important;
+  flex: 1 1 auto;
 }
 
-.active-effect-sheet textarea {
+.script-config textarea {
```
moo-man, line 14774: The real pain I was experiencing with the script window is making sure it works with both normal scripts and dialog scripts (which have 3 other subscript fields).

modules/system/overrides.js
silentmark: changed

modules/actor/actor-wfrp4e.js
silentmark: changed

static/css/wfrp4e.css
silentmark: I will play around with that.

modules/apps/script-config.js
```diff
@@ -46,7 +46,26 @@ export default class ScriptConfig extends FormApplication
...
+                esVersion: 13,
+            }));
+            this.editor.session.on('changeMode', function(e, session){
+                if ("ace/mode/javascript" === session.getMode().$id) {
+                    if (!!session.$worker) {
+                        session.$worker.send("setOptions", [{
+                            "esversion": 13
+                        }]);
+                    }
+                }
```
moo-man, line 64: What is this doing exactly?

static/css/wfrp4e.css
moo-man: Testing it, changing to a dialog effect simply disables the ace editor, which I think is fine. Feel free to tinker but comment when you're satisfied and want me to merge.

modules/apps/script-config.js
moo-man: This should be good to merge but I'm still curious about this.

silentmark: this should set esversion to minimize the number of syntax errors for newer javascript, but i'm afraid that it will not work, as the ace library module is using v1.10 while the newest one is v1.32 anyway it seems to look good: 
image

moo-man: > but i'm afraid that it will not work
Is this needed then? Or am I misunderstanding.

silentmark: it will do no harm and in future can be beneficial

PR #1851, opened 2023-12-16
Disease code refactor (Part 1)

Forien:
Refactored some of the code responsible for handling Actor's diseases. 

For the most part, changes could be described as one of the following:
- Styling improvements
- Reducing code duplication (DRY)
- Improving readability and adding JSDoc comments
- Replacing `.then()` syntax with `await` syntax inside `async` function

But there are also several instances of actual changes and improvements, such as:
- Adding `await` in places, where lack of it could lead to data loss and/or race conditions
- Fixing actual errors.

---
**More info in commit [review/comments](https://github.com/moo-man/WFRP4e-FoundryVTT/pull/1851/files).**

---
I realize that **most of these methods** will need an additional pass when Disease are migrated (currently, I am unable to created new Items on this branch because of migration errors). 

However, there are some lines that need an additional scrutiny:

~~### 1. Here, both "diseases" are not found in the compendiums, so `lingeringDisease` will always be `null`.~~
~~https://github.com/moo-man/WFRP4e-FoundryVTT/pull/1851/files#diff-c588f96685c7089ee5b43e41afbdb538af2993de610f58860343fe75a30cab2eR1414-R1418~~
Resolved in commits.

### 2. Here, the variable `removeEffects` was never defined (and if it were, it would already be an array, so the line itself is somewhat weird). 
https://github.com/moo-man/WFRP4e-FoundryVTT/pull/1851/files#diff-c588f96685c7089ee5b43e41afbdb538af2993de610f58860343fe75a30cab2eR1426

### DIFF
--- a/modules/actor/actor-wfrp4e.js
+++ b/modules/actor/actor-wfrp4e.js
@@ -1283,125 +1283,155 @@ export default class ActorWfrp4e extends WFRP4eDocumentMixin(Actor)
     }
   }
 
+  /**
+   * Decrements all injuries of this Actor
+   *
+   * @return {Promise<void>}
+   */
   async decrementInjuries() {
-    this.injuries.forEach(i => this.decrementInjury(i))
+    for (let i of this.injuries)
+      await this.decrementInjury(i);
   }
 
+  /**
+   * Decrements a `duration` value for the Injury.
+   * If `duration` reaches 0, inform about Injury being finished.
+   *
+   * @param {ItemWfrp4e}injury
+   *
+   * @return {Promise<*>}
+   */
   async decrementInjury(injury) {
     if (isNaN(injury.system.duration.value))
       return ui.notifications.notify(game.i18n.format("CHAT.InjuryError", { injury: injury.name }))
 
-    injury = duplicate(injury)
-    injury.system.duration.value--
+    injury = foundry.utils.duplicate(injury);
+    injury.system.duration.value--;
 
-    if (injury.system.duration.value < 0)
+    if (injury.system.duration.value <= 0) {
       injury.system.duration.value = 0;
-
-    if (injury.system.duration.value == 0) {
-      let chatData = game.wfrp4e.utility.chatDataSetup(game.i18n.format("CHAT.InjuryFinish", { injury: injury.name }), "gmroll")
-      chatData.speaker = { alias: this.name }
-      ChatMessage.create(chatData)
+      let chatData = game.wfrp4e.utility.chatDataSetup(game.i18n.format("CHAT.InjuryFinish", { injury: injury.name }), "gmroll");
+      chatData.speaker = { alias: this.name };
+      ChatMessage.create(chatData);
     }
-    this.updateEmbeddedDocuments("Item", [injury]);
+
+    await this.updateEmbeddedDocuments("Item", [injury]);
   }
 
-
+  /**
+   * Decrements all diseases of this Actor
+   *
+   * @return {Promise<void>}
+   */
   async decrementDiseases() {
-    this.diseases.forEach(d => this.decrementDisease(d))
+    for (let d of this.diseases)
+      await this.decrementDisease(d);
   }
 
+  /**
+   * Decrements a value for Disease's `incubation` or `duration` attribute, depending on whether it is active or not.
+   * If value reaches 0, activates (if `incubation`), or finishes (if `duration`) the Disease.
+   *
+   * @param {ItemWfrp4e} disease
+   *
+   * @return {Promise<void>}
+   */
   async decrementDisease(disease) {
-    let d = duplicate(disease)
-    if (!d.system.duration.active) {
-      if (Number.isNumeric(d.system.incubation.value)) {
+    let d = foundry.utils.duplicate(disease);
+    let type = d.system.duration.active ? 'duration' : 'incubation';
 
-        d.system.incubation.value--
-        if (d.system.incubation.value <= 0) {
-          this.activateDisease(d)
-          d.system.incubation.value = 0;
-        }
-      }
-      else {
-        let chatData = game.wfrp4e.utility.chatDataSetup(`Attempted to decrement ${d.name} incubation but value is non-numeric`, "gmroll", false)
-        chatData.speaker = { alias: this.name }
-        ChatMessage.create(chatData)
-      }
-    }
-    else {
-      if (Number.isNumeric(d.system.duration.value)) {
+    if (Number.isNumeric(d.system[type].value)) {
+      d.system[type].value--;
 
-        d.system.duration.value--
-        if (d.system.duration.value == 0)
-          this.finishDisease(d)
+      if (d.system[type].value <= 0) {
+        d.system[type].value = 0;
+
+        if (type === 'incubation')
+          await actor.activateDisease(d);
+
+        if (type === 'duration')
+          await actor.finishDisease(d);
       }
-      else {
-        let chatData = game.wfrp4e.utility.chatDataSetup(`Attempted to decrement ${d.name} duration but value is non-numeric`, "gmroll", false)
-        chatData.speaker = { alias: this.name }
-        ChatMessage.create(chatData)
-      }
+    } else {
+      let chatData = game.wfrp4e.utility.chatDataSetup(`Attempted to decrement ${d.name} ${type} but value is non-numeric`, "gmroll", false);
+      chatData.speaker = {alias: this.name};
+      ChatMessage.create(chatData);
     }
-    this.updateEmbeddedDocuments("Item", [d])
+
+    await this.updateEmbeddedDocuments("Item", [d])
   }
 
+  /**
+   *
+   * @param {ItemWfrp4e} disease
+   *
+   * @return {Promise<void>}
+   */
   async activateDisease(disease) {
     disease.system.duration.active = true;
     disease.system.incubation.value = 0;
-    let msg = game.i18n.format("CHAT.DiseaseIncubation", { disease: disease.name })
+    let msg = game.i18n.format("CHAT.DiseaseIncubation", { disease: disease.name });
+
     try {
-      let durationRoll = (await new Roll(disease.system.duration.value).roll()).total
-      msg += game.i18n.format("CHAT.DiseaseDuration", { duration: durationRoll, unit: disease.system.duration.unit })
+      let durationRoll = (await new Roll(disease.system.duration.value).roll()).total;
+      msg += game.i18n.format("CHAT.DiseaseDuration", { duration: durationRoll, unit: disease.system.duration.unit });
       disease.system.duration.value = durationRoll;
-    }
-    catch (e) {
-      msg += game.i18n.localize("CHAT.DiseaseDurationError")
+    } catch (e) {
+      msg += game.i18n.localize("CHAT.DiseaseDurationError");
     }
 
-    let chatData = game.wfrp4e.utility.chatDataSetup(msg, "gmroll", false)
-    chatData.speaker = { alias: this.name }
-    ChatMessage.create(chatData)
+    let chatData = game.wfrp4e.utility.chatDataSetup(msg, "gmroll", false);
+    chatData.speaker = { alias: this.name };
+    ChatMessage.create(chatData);
   }
 
+  /**
+   * Finishes disease and handles Lingering symptoms.
+   *
+   * @param {ItemWfrp4e} disease
+   *
+   * @return {Promise<void>}
+   */
   async finishDisease(disease) {
-
     let msg = game.i18n.format("CHAT.DiseaseFinish", { disease: disease.name })
 
-    if (disease.system.symptoms.includes("lingering")) {
+    if (disease.system.symptoms.value.includes("lingering")) {
       let lingering = disease.effects.find(e => e.name.includes("Lingering"))
       if (lingering) {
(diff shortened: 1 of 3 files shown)


modules/actor/actor-wfrp4e.js
```diff
@@ -1283,125 +1283,155 @@ export default class ActorWfrp4e extends WFRP4eDocumentMixin(Actor)
...
   }
 
+  /**
+   * Decrements all injuries of this Actor
+   *
+   * @return {Promise<void>}
+   */
   async decrementInjuries() {
-    this.injuries.forEach(i => this.decrementInjury(i))
+    for (let i of this.injuries)
```
Forien, line 1292: Changing `forEach` to `for…of` allows async function to properly `await` all execution

modules/actor/actor-wfrp4e.js
```diff
@@ -1283,125 +1283,155 @@ export default class ActorWfrp4e extends WFRP4eDocumentMixin(Actor)
...
+   *
+   * @return {Promise<*>}
+   */
   async decrementInjury(injury) {
     if (isNaN(injury.system.duration.value))
       return ui.notifications.notify(game.i18n.format("CHAT.InjuryError", { injury: injury.name }))
 
-    injury = duplicate(injury)
-    injury.system.duration.value--
+    injury = foundry.utils.duplicate(injury);
```
Forien, line 1308: Just in case there will be a time when Foundry will stop supporting global functions. They seem to push usage of the `foundry.utils` getter, and it makes for better IDE highlighting.

modules/actor/actor-wfrp4e.js
```diff
@@ -1283,125 +1283,155 @@ export default class ActorWfrp4e extends WFRP4eDocumentMixin(Actor)
...
+    injury.system.duration.value--;
 
-    if (injury.system.duration.value < 0)
+    if (injury.system.duration.value <= 0) {
       injury.system.duration.value = 0;
-
-    if (injury.system.duration.value == 0) {
```
Forien, line 1300: Swapping two `if` statements into single one, that in the end works exactly the same

modules/actor/actor-wfrp4e.js
```diff
@@ -1283,125 +1283,155 @@ export default class ActorWfrp4e extends WFRP4eDocumentMixin(Actor)
...
-      chatData.speaker = { alias: this.name }
-      ChatMessage.create(chatData)
+      let chatData = game.wfrp4e.utility.chatDataSetup(game.i18n.format("CHAT.InjuryFinish", { injury: injury.name }), "gmroll");
+      chatData.speaker = { alias: this.name };
+      ChatMessage.create(chatData);
     }
-    this.updateEmbeddedDocuments("Item", [injury]);
-  }
 
+    await this.updateEmbeddedDocuments("Item", [injury]);
```
Forien, line 1318: adding `await`, since it's an `async` function. Just to avoid potential race condition

modules/actor/actor-wfrp4e.js
```diff
@@ -1283,125 +1283,155 @@ export default class ActorWfrp4e extends WFRP4eDocumentMixin(Actor)
...
+  }
 
+  /**
+   * Decrements all diseases of this Actor
+   *
+   * @return {Promise<void>}
+   */
   async decrementDiseases() {
-    this.diseases.forEach(d => this.decrementDisease(d))
+    for (let d of this.diseases)
```
Forien, line 1327: Changing `forEach` to `for…of` allows async function to properly `await` all execution

modules/actor/actor-wfrp4e.js
```diff
@@ -1283,125 +1283,155 @@ export default class ActorWfrp4e extends WFRP4eDocumentMixin(Actor)
...
-      else {
-        let chatData = game.wfrp4e.utility.chatDataSetup(`Attempted to decrement ${d.name} incubation but value is non-numeric`, "gmroll", false)
-        chatData.speaker = { alias: this.name }
-        ChatMessage.create(chatData)
-      }
-    }
-    else {
-      if (Number.isNumeric(d.system.duration.value)) {
+    let d = foundry.utils.duplicate(disease);
+    let type = d.system.duration.active ? 'duration' : 'incubation';
```
Forien, line 1342: Simplified an `if` statement that practically reused the same code in both `if` and `else` parts.

modules/actor/actor-wfrp4e.js
```diff
@@ -1283,125 +1283,155 @@ export default class ActorWfrp4e extends WFRP4eDocumentMixin(Actor)
...
-        chatData.speaker = { alias: this.name }
-        ChatMessage.create(chatData)
+    if (Number.isNumeric(d.system[type].value)) {
+      d.system[type].value--;
+
+      if (d.system[type].value <= 0) {
+        d.system[type].value = 0;
+
+        if (type === 'incubation')
+          await actor.activateDisease(d);
```
Forien, line 1350: Added `await`, since we are updating the disease further in the same function, so we don't lose the data.

modules/actor/actor-wfrp4e.js
```diff
@@ -1283,125 +1283,155 @@ export default class ActorWfrp4e extends WFRP4eDocumentMixin(Actor)
...
+
+        if (type === 'duration')
+          await actor.finishDisease(d);
```
Forien, line 1353: Added `await`, since we are updating the disease further in the same function, so we don't lose the data.

modules/actor/actor-wfrp4e.js
```diff
@@ -1283,125 +1283,155 @@ export default class ActorWfrp4e extends WFRP4eDocumentMixin(Actor)
...
       }
+    } else {
+      let chatData = game.wfrp4e.utility.chatDataSetup(`Attempted to decrement ${d.name} ${type} but value is non-numeric`, "gmroll", false);
+      chatData.speaker = {alias: this.name};
+      ChatMessage.create(chatData);
     }
-    this.updateEmbeddedDocuments("Item", [d])
+
+    await this.updateEmbeddedDocuments("Item", [d])
```
Forien, line 1361: Added `await`, since it's an `async` function

modules/actor/actor-wfrp4e.js
```diff
@@ -1283,125 +1283,155 @@ export default class ActorWfrp4e extends WFRP4eDocumentMixin(Actor)
...
-              fromUuid("Compendium.wfrp4e-core.diseases.kKccDTGzWzSXCBOb").then(disease => {
-                this.createEmbeddedDocuments("Item", [disease.toObject()])
-              })
-            }
-            else if (negSL >= 6) {
-              msg += game.i18n.localize("CHAT.LingeringRot")
-              fromUuid("Compendium.wfrp4e-core.diseases.M8XyRs9DN12XsFTQ").then(disease => {
-                this.createEmbeddedDocuments("Item", [disease.toObject()])
-              })
-            }
```
Forien, line 1391: In this code block I did two things.

1. I replaced the `.then()` syntax with the `await` syntax, since we are in `async` function and it improves readability of the code
2. I removed the call to deprecated `actor.basicTest(setupData)` in favor of `test.roll()`

**Probably worth mentioning that both UUIDs provided here return `null`, so it needs fixing.**

modules/actor/actor-wfrp4e.js
```diff
@@ -1283,125 +1283,155 @@ export default class ActorWfrp4e extends WFRP4eDocumentMixin(Actor)
...
+   * @param {ItemWfrp4e} disease
+   *
+   * @return {Promise<void>}
+   */
   async finishDisease(disease) {
-
     let msg = game.i18n.format("CHAT.DiseaseFinish", { disease: disease.name })
 
-    if (disease.system.symptoms.includes("lingering")) {
+    if (disease.system.symptoms.value.includes("lingering")) {
```
Forien, line 1398: Fixing error that happened every time:
image

modules/actor/actor-wfrp4e.js
```diff
@@ -1283,125 +1283,155 @@ export default class ActorWfrp4e extends WFRP4eDocumentMixin(Actor)
...
+            lingeringDisease = await fromUuid("Compendium.wfrp4e-core.diseases.M8XyRs9DN12XsFTQ");
           }
-        }))
+
+          if (lingeringDisease)
+            await this.createEmbeddedDocuments("Item", [lingeringDisease.toObject()]);
+        }
       }
+    } else {
+      // await this.deleteEmbeddedDocuments("ActiveEffect", [removeEffects]);
```
Forien, line 1426: `removeEffects` was not defined at all in the function, so commenting it for now. I realize that these functions will need an additional pass once Diseases and Symptoms are migrated.

modules/actor/actor-wfrp4e.js
moo-man: They also seem to push `deepClone` instead of `duplicate` so that might as well change too.

modules/actor/actor-wfrp4e.js
moo-man: Might be worth changing `decrementDisease` to return the modified object data, then calling `updateEmbeddedDocuments` to get it all done in a single database call, but that might need to wait until diseases utilize the new effect system.

modules/actor/actor-wfrp4e.js
```diff
@@ -1283,125 +1283,155 @@ export default class ActorWfrp4e extends WFRP4eDocumentMixin(Actor)
...
-            else if (negSL >= 6) {
-              msg += game.i18n.localize("CHAT.LingeringRot")
-              fromUuid("Compendium.wfrp4e-core.diseases.M8XyRs9DN12XsFTQ").then(disease => {
-                this.createEmbeddedDocuments("Item", [disease.toObject()])
-              })
-            }
+        let difficulty = lingering.name.substring(lingering.name.indexOf("(") + 1, lingering.name.indexOf(")")).toLowerCase();
+
+        let test = await this.setupSkill(game.i18n.localize("NAME.Endurance"), { difficulty });
+        test.roll();
```
moo-man, line 1404: `test.roll()` is async, does this work without awaiting it? Probably should await just in case.

modules/actor/actor-wfrp4e.js
```diff
@@ -1283,125 +1283,155 @@ export default class ActorWfrp4e extends WFRP4eDocumentMixin(Actor)
...
+
+          if (negSL <= 1) {
+            let roll = (await new Roll("1d10").roll()).total;
+            msg += game.i18n.format("CHAT.LingeringExtended", { duration: roll });
+          } else if (negSL <= 5) {
+            msg += game.i18n.localize("CHAT.LingeringFestering");
+            lingeringDisease = await fromUuid("Compendium.wfrp4e-core.diseases.kKccDTGzWzSXCBOb");
+          } else if (negSL >= 6) {
+            msg += game.i18n.localize("CHAT.LingeringRot");
+            lingeringDisease = await fromUuid("Compendium.wfrp4e-core.diseases.M8XyRs9DN12XsFTQ");
```
moo-man, line 1418: Regarding `null`, change `diseases` to `items` and it should find them.

modules/actor/actor-wfrp4e.js
Forien: Yeah, you're right :D

modules/actor/actor-wfrp4e.js
Forien: That might be worth looking into, however for now I just wanted to fix the obvious errors I was getting and prepare code to be more easily maintained in the future. I did it for this branch because I know that this will most likely need more scrutiny when the new effects are done.

modules/actor/actor-wfrp4e.js
Forien: Changed

modules/model/item/disease.js
```diff
@@ -1,59 +1,67 @@
...
-      {
-        let properties = [];
-        properties.push(`<b>${game.i18n.localize("Contraction")}:</b> ${this.contraction.value}`);
-        properties.push(`<b>${game.i18n.localize("Incubation")}:</b> <a class = 'chat-roll'><i class='fas fa-dice'></i> ${this.incubation.value}</a>`);
-        properties.push(`<b>${game.i18n.localize("Duration")}:</b> <a class = 'chat-roll'><i class='fas fa-dice'></i> ${this.duration.value}</a>`);
-        properties.push(`<b>${game.i18n.localize("Symptoms")}:</b> ${(this.symptoms.value.split(",").map(i => i = "<a class ='symptom-tag'><i class='fas fa-user-injured'></i> " + i.trim() + "</a>")).join(", ")}`);
-        if (this.permanent.value)
-          properties.push(`<b>${game.i18n.localize("Permanent")}:</b> ${this.permanent.value}`);
-        return properties;
-      }
```
Forien, line 58: Whole file shows as changed becuase of formatting, actual change can be seen in specific commit: https://github.com/moo-man/WFRP4e-FoundryVTT/pull/1851/commits/ed6fe9651f0bf56b074e4b23d1f8db0e2e7ceee7

Forien:
New code will automatically set `incubation.value` to `0` in disease contracted from **Lingering** symptom.

I would suggest changing the `Festering Wound` (`Compendium.wfrp4e-core.items.Item.kKccDTGzWzSXCBOb`) incubation to `1d10` and `days` as unit. Currently unit is missing and `incubation.value` is not parseable.
I understand the need of clarification, but I think it should be moved to description.

Alternatively, we could modify the code to only try to parse stuff **before a comma** as `roll formula`.

PR #1853, opened 2023-12-19
Pass actual ActorWfrp4e instance to scripts 

Forien:
Right now, on effects such as `Prepare Data`, the `args.actor` holds Actor's Data Model. This breaks some effects that rely on this, such as `Malaise` that uses `args.actor.hasCondition()` and `args.actor.addCondition()`.

Proposed PR will change it instead to Actor's actual instance so effect scripts can have unchanged access to the entire Actor Document.

### DIFF
--- a/modules/model/actor/standard.js
+++ b/modules/model/actor/standard.js
@@ -94,11 +94,11 @@ export class StandardActorModel extends BaseActorModel {
         flags.resolute = 0
         flags.ambi = 0;
 
-        this.parent.runScripts("prePrepareData", { actor: this })
+        this.parent.runScripts("prePrepareData", { actor: this.parent })
     }
 
     computeDerived() {
-        this.parent.runScripts("prePrepareItems", {actor : this})
+        this.parent.runScripts("prePrepareItems", {actor : this.parent })
         this.computeItems();
         super.computeDerived();
         // Recompute bonuses as active effects may have changed it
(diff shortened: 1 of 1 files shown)


moo-man:
Iffy on this, though mainly because I want to keep as much congruity as I can with how I implemented it in IM (which doesn't pass the parent). 

I would need to see how big of an impact this has, i.e. if this doesn't get changed, how many scripts are affected.

Forien:
Unless I'm mistaken, all `prepareData` and similar effects will be affected and need a rewrite, because even a simple:
```
args.actor.system.characteristics -= 10;
```
Is no longer valid, since the Data Model is already the `system` and would need script like this:
```
args.actor.characteristics -= 10;
```

I feel like changing it from Document to Data Model is huge change that breaks compatibility with almost everything that does effects using any of:

- prePrepareData
- prepareData
- prePrepareItems
- preWoundCalc
- woundCalc

Additionally, I think it's bad, because it would bring a lot of confusion from people asking:
> Why sometimes in effects `args.actor` is actor and sometimes it's not?

A compromise could be providing both, or (if you really are against passing Document instance), changing argument's name from `actor` to `model` to remove confusion. 
_(and you could always update IM too if you want to reuse components...)_

moo-man:
> you could always update IM too 

Well, the effects have all already been written. Nothing for WFRP4e has been rewritten yet for the new effects. It looks like changing this would prevent needing to rewrite many effects though so it's worth considering.

Forien:
Could always add `model` as additional arg ;)

moo-man:
🤷 what's the point, `args.actor.system` gets you there.

PR #1857, opened 2023-12-28
fixes to migration and applicable traits

Forien:


modules/model/item/trait.js
```diff
@@ -164,4 +164,10 @@ export class TraitModel extends PropertiesMixin(BaseItemModel)
       }
 
 
+    shouldTransferEffect(effect)
+    {
+      let isExcluded = this.parent.actor?.excludedTraits?.includes(this.parent._id) ?? false;
+      return isExcluded === false;
+    }
```
moo-man, line 171: change this to `return !this.parent.actor?.excludedTraits?.includes(this.id)` and I will merge.

PR #1858, opened 2023-12-29
Chargen extension to allow other random talent tables

Forien:
## The Problem
Before, there was only single Random Talents table existing and it was hard coded. This made it very difficult to come up with creative ways of defining new, fan-made species that could have "3 Random Talents from Species Table". 

## Solution
Changes provided in this Pull Request extend the CharGen application in such a way, that "Random Talents" are now sorted by tables' keys, allowing for very high degree of customization of random talents for a species, including ability to define random talents from multiple different tables to be rolled for the same species at the same time. 

## How?
When defining a new Species, the following should be used to define the random talents:
```js
config.speciesRandomTalents["human"] = {
  talents: 3
};
``` 
`talents` in this example is the `key` of default Random Talents Rollable Table.

Defining a species that rolls from both default table and new, custom one would look like this:
```js
config.speciesRandomTalents["human"] = {
  talents: 1,
  'tilean-talents': 2
};
```

### What about subspecies?
This works also on the subspecies level and can be defined by providing the object to the `randomTalents` propery of the subspecies. It would look like this:
```js
config.subspecies['human']['tilean'] = {
  randomTalents: {
    talents: 1,
    'tilean-talents': 2
  }
}
 ```

### Random optional talents from specific table?
Yes, it's possible using extended version of `random[x]` format, that is `random[x][key]`.   
Consider the following definition:
```js
config.speciesTalents["human"] = {
  "Doomed",
  "Read/Write, random[1]",
  "Ambidextrous, random[2][tilean-talents]"
};
```
Player would be able to chose between `Read/Write` or `1 Additional Random Talent` rolled from default table, and also between `Ambidextrous` or `2 Additional Random Talents` from table with `tilean-talents` key. 

## What if defined table does not exist?
Errors will appear for the user, but otherwise CharGen will keep working as if the random talents of that table were not defined.

## Backwards compatibility?
If there is a numerical value on the `speciesTalents` or `subspecies.talents`, it will be assumed to be describing amount of random talents from the default table, no change is required for currently existing (sub)species extensions.

## Other minor changes
- Fixed the UUIDs for opening Journals about Character Creation
- Expanded ability of dynamically defining new Stage

---

### TODO list:
- [x] Add structure allowing to contain talent data from multiple tables
- [x] Handle and maintain backwards-compatibility
- [x] Allow rolling random talents from any table by key
- [x] Accurately mark duplicates
- [x] Allow rerolling duplicates
- [x] Accurately detect if chosen talent caused rolled random one to become a duplicate
- [x] Properly handle `random[x]` syntax across table data
- [x] Allow `random[x]` to additionally (and optionally) determine table (`random[x,key]` or `random[x][key]`). If no `key` is specified, default to `talents`.
- [x] Properly remove talent from relevant table if `random[x]` talent is no longer chosen
- [x] Properly add all talents to character data on completing the stage
- [x] Test the living Dhar out of it with as many various cases I can think of, including unofficial ones
- [x] Cleanup code of all artifacts and `@todo` tags

### Optionally:
- [x] Fix **UUIDs** for `JournalEntryPage` for every CharGen step
- [x] Determine if styling needs adjusting to handle talent lists and buttons from multiple tables


modules/apps/chargen/skills-talents.js
```diff
@@ -208,6 +283,8 @@ export class SkillsTalentsStage extends ChargenStage {
       valid = false
     }
 
+    // Should this allow partial allocation? I think it should not validate if allocated value is anything other than 40
+    // simply because it could be done by accident and hard to retrace in such a case. /Forien
```
Forien, line 287: @moo-man Is it intentional to allow submitting the stage if not all points were allocated?

Forien: Same with skills, is it intentional to allow submitting stage if less than 3 skills have been selected for each advancement "tier"?

moo-man: I didn't have any specific reasons for validating this or not validating that, just what seemed important/easy to mess up.

modules/apps/chargen/char-gen.js
```diff
@@ -340,19 +345,17 @@ export default class CharGenWfrp4e extends FormApplication {
...
-  addStage(stage, index)
-  {
-    let stageObj = stage.stageData()
-    if (index == undefined)
-    {
+  addStage(stage, index, stageData = {}) {
+    let stageObj = stage.stageData();
+    stageObj = foundry.utils.mergeObject(stageObj, stageData);
+
+    if (index === undefined) {
```
moo-man, line 352: Would it be worth simply moving all stageData into the stage itself so this isn't needed?

modules/apps/chargen/skills-talents.js
```diff
@@ -54,96 +70,95 @@ export class SkillsTalentsStage extends ChargenStage {
...
   context = {
     speciesSkills: {},
     speciesTalents: {
       normal: [],
       chosen: [],
-      random: [],
-      rolled: false,
-      randomCount: 0,
       choices: [],
+      tables: new Map()
```
moo-man, line 83: Is there a particular reason this is a Map instead of just an object?

Forien: I think I initially had some more specific reason in mind, but now that I'm trying to think about it... it's just there because I started with it. 

I guess it's cleaner in my eyes to do array conversion with
```js
[...variable.values()]
``` than with
```js
[...Object.values(variable)]
```

But yeah, I can't think of any actual argument to use it over the object, and the same would be the other way around, I couldn't argue to use object over Map... 
Performance in this case wouldn't matter anyway, so that argument for `Map` is not valid. 
So I guess I'm left with "I like Maps".

moo-man: I guess I'm simply not sure why you're using the spread operator in both cases because I'm not sure there's a point to it?

modules/apps/chargen/char-gen.js
Forien: The Stage itself has its data hardcoded in the `static stageData()` method, the potential issue is that it's the same, always.

The main reason for that change is to allow to reuse the same Stage Class with different data (for example different `name` etc) and also to provide more data to the Stage. 

And since to add new stage we don't initialize the stage, but only pass a class, I don't know how else easily do this.

Unless we would switch to pass an already instantiated stage object instead of a class to the `addStage()`?

I'm curious what you have in mind

modules/apps/chargen/skills-talents.js
Forien: Unless I'm mistaken. the spread operator `[...someMap.values()]` is the same as calling `Array.from(someMap.values())`, just looks nicer to me with less brackets etc. As far as I know they are synonymous with each other? 🤔 

As for why converting to array... it's because while Map/Object makes it easier/cleaner to store and keep values properly sorted by unique key etc (for the purpose of adding and removing talents from proper table without having to do `.find()` shenanigans), array makes it easier to do bulk operations. Like checking for duplicates, or reducing all random talents to single array for submission.

moo-man: Sorry I didn't check and thought `map.values()` returned an Array. `Object.values()` called on an object does return an array, which is why I thought it would be cleaner.

modules/apps/chargen/char-gen.js
moo-man: > reuse the same Stage Class with different data (for example different name etc) 

I'm not sure I follow this, when would this ever be needed? Do you have an example? 

> And since to add new stage we don't initialize the stage, but only pass a class, I don't know how else easily do this.

Well, `stageData` is static, we wouldn't need to have an instance, unless I'm misunderstanding something. 

I'm not really leaning either way I just wonder if it'd be simpler to move the stageData into the class as static data, though I don't know what we would do with the `app` property.

modules/apps/chargen/skills-talents.js
moo-man: It's not a huge deal, it's just a weird outlier that bothers me for only that reason. If it's easy to change to an object go ahead, but if it causes issues don't worry about it.

modules/apps/chargen/char-gen.js
Forien: The only thing I actually have in mind is making a stage that allows to select items to start with. It could be used to pickup petty spells, arcane spells, custom fanmade spells, maybe some blessings or something.

I thought of adding something like that in my Armoury, mainly to pick the Runebound Spells, but then I realized that once I do it, then the exact same class could be used to allow players selecting their starting spells and maybe other things. All that would need to change is the title and maybe a callback or a flag that would control the possible selection. 

But then again, I guess I could just do one class and then extend it for every use case with absolute minimal required overrides... 

I didn't really think it super through, as I don't have immediate need for this, but what I commited seemed like the easiest thing to achieve that with as little change as possible. 

We can put it on hold for now if you prefer to sit on it

moo-man: Well I think that would simply be a case of subclassing some `SpellSelectionStage` and redefining its data.

Regardless, I don't think it's in the scope of this PR and can be changed later if needed.

.github/workflows/release.yml
```diff
@@ -46,18 +46,46 @@ jobs:
...
-      - name: Generate Changelog
-        id: changelog
-        uses: mikepenz/release-changelog-builder-action@v2
-        env:
-          GITHUB_TOKEN: ${{ secrets.PAT }}
+      # - name: Generate Changelog
+      #   id: changelog
+      #   uses: mikepenz/release-changelog-builder-action@v2
+      #   env:
+      #     GITHUB_TOKEN: ${{ secrets.PAT }}
```
moo-man, line 83: I'm trying to figure out why these changes are showing (as well as others) when I've merged master into `effect-refactor` already.

Forien: Not sure if this had something to do with it? 
image

Because after my last commit I didn't see there changes from other commits 🤔

Forien: I went ahead and merged current `effect-refactor` into my branch, which seem to fix the `files changed` tab.

moo-man: I switched the branches because that has fixed the issue before. But it looks good now.

PR #1860, opened 2023-12-31
More JSDoc data to help my IDE

Forien:


modules/actor/mixin.js
```diff
@@ -195,4 +201,6 @@ export default WFRP4eDocumentMixin = (cls) => class extends cls {
         }
     }
 
-}
\ No newline at end of file
+}
+
+export default WFRP4eDocumentMixin;
```
moo-man, line 206: Does separating this help JSDoc in any way?

Forien: No, but it removes the IDE's `Unresolved variable` error. I can revert it as it doesn't stop the code from executing. But then again, this header would work just as fine:
```js
export default (cls) => class extends cls {
```
I don't really have a preference on this, I'm fine either way

image

PR #1865, opened 2024-01-08
Fix Enter not working when editing effect scripts and/or macros

Forien:
The issue when editing macros/effect scripts was because event listener for `Enter` was bound to Roll Dialog.

Last few days I was redoing Runes in my module, and that involves hours and hours of modifying script effects, testing etc. 
Normally `Enter` works properly and makes new line. However, during these few days, on several ocassions (like 4 or 5 times?), Enter stopped working in script Editor.

Today I finally was able to figure it out. (console helps, duh)

For whatever reason, **sometimes**, the Roll Dialog is not properly closed and its event listener is not being removed, so it is always listening, meaning that whenever user is focused on an element that doesn't override the keypress (so not chat window, canvas, form application etc, but textareas like Macro editor or Script Effect editor), the following error would appear in the console and would prevent new line from forming:

image

It's very hard to reproduce (took me 30+ minutes of doing opposed tests in multiple ways to get that screenshot). And once I got it, I couldn't easily reproduce it again, so it seems random. 

Error points to `data.targets = Array.from(data.targets).map(t => t.actor.speakerData(t.document));`, but if you consider what would happen if this line happens more than once, it's clear that `t.document` would not exist in `data.targets` once it's converted to `speakerData`. But I don't think this is the issue, I think it's that randomly the Event Listener is not removed and Roll Dialog being submitted twice.

## Solution
I added two, somewhat reduntant checks. Now `Enter` will only submit Roll Dialog, if Roll Dialog: is rendered **and** is the Active Window. 

Since it's sort of a stop gap solution, it might not help with Event Listener randomly not being removed, but should mitigate any errors that this could lead to. 

Side effect (positive in my opinion) is that with Roll Dialog opened, `Enter` won't submit it if pressed while focus is on another Window

### DIFF
--- a/modules/apps/roll-dialog/roll-dialog.js
+++ b/modules/apps/roll-dialog/roll-dialog.js
@@ -373,7 +373,10 @@ export default class RollDialog extends Application {
 
     _onKeyPress(ev)
     {
-        if (ev.key == "Enter")
+        if (!this.rendered) return;
+        if (this !== ui.activeWindow) return;
+
+        if (ev.key === "Enter")
         {
             this.submit(ev); 
         }
--- a/modules/apps/roll-dialog/roll-dialog.js
+++ b/modules/apps/roll-dialog/roll-dialog.js
@@ -376,7 +376,7 @@ export default class RollDialog extends Application {
         if (!this.rendered) return;
         if (this !== ui.activeWindow) return;
 
-        if (ev.key === "Enter")
+        if (ev.key == "Enter")
         {
             this.submit(ev); 
         }


Forien:
Was able to figure how to reliably reproduce the error

1. When using Weapon to perform an Attack Test, when I change difficulty to +60 and then submit, then `Enter` stops working. But no error in console
2. However, if I do the same, but **with target**, then also error appears

Steps:
1. Put 2 Tokens (Token A and Token B) on the canvas, give them hand weapons
2. Have Token A roll a Weapon Test without targetting
3. Change difficulty in Dialog to +60
4. Submit roll
5. Open Macro editor and try making new line (observe no error in console)
6.  Target Token B
7. Have Token A perform another Weapon Test, this time opposing one
8. Again, change difficulty to +60 and submit
9. Try to make new line in Editor, observe that in addition to no new line, an error appears in console. 

Finishing the opposing test or marking it as unopposed bears no effect on the bug.

moo-man:
Went with a different fix that I believe solves the issue, see: https://github.com/moo-man/WFRP4e-FoundryVTT/commit/8902a106f1cb71cc53342e24077cc40f7dfdd224

Forien:
> Went with a different fix that I believe solves the issue, see: [8902a10](https://github.com/moo-man/WFRP4e-FoundryVTT/commit/8902a106f1cb71cc53342e24077cc40f7dfdd224)

Should help, but I would still argue that stopping the listener from activating when window isn't focused would be a good idea.

For example someone opens the roll dialog, but for whatever reason they need to edit something else and press enter while on different app. Yes, I am aware it's a not very likely edge case

moo-man:
I think it goes both ways, if some effect applied to multiple actors causes a bunch of roll dialogs to show up, I'd like enter to simply roll them all, but that is also very edge case.

Forien:
> I think it goes both ways, if some effect applied to multiple actors causes a bunch of roll dialogs to show up, I'd like enter to simply roll them all, but that is also very edge case.

Oh right... 🤔 didn't think about that. 

Yeah I guess your case is more likely than mine anyway, so I guess it's good enough :)

PR #1875, opened 2024-01-17
Multiple changes to support Item SubTypes

Forien:
PR with changes I need in order to properly add my own Item SubTypes.

Includes several fixes to minor errors that would arise with SubTypes, as well as changes to better support them, including better handling of Item Sheets, Properties etc.

Also includes several changes I need to be able to cast spells that aren't on actors, using custom class that extends the CastTest.


modules/apps/item-properties.js
```diff
@@ -13,28 +13,50 @@ export default class ItemProperties extends FormApplication {
...
-            this.qualities = {}
+        super(...args);
+
+        if (ItemProperties.hasWeaponProperties(this.object)) {
+            this.qualities = foundry.utils.deepClone(game.wfrp4e.config.weaponQualities);
+            this.flaws = foundry.utils.deepClone(game.wfrp4e.config.weaponFlaws);
+        } else if (ItemProperties.hasArmourProperties(this.object)) {
+            this.qualities = foundry.utils.deepClone(game.wfrp4e.config.armorQualities);
+            this.flaws = foundry.utils.deepClone(game.wfrp4e.config.armorFlaws);
+        } else {
```
Forien, line 24: not only `trapping`, but every other type should be handled. 

threw errors otherwise

modules/model/item/components/overcast.js
```diff
@@ -27,7 +27,14 @@ export class OvercastItemModel extends BaseItemModel {
               additional  : new fields.StringField(),
               characteristic  : new fields.StringField(),
               bonus : new fields.BooleanField(),
-         }),
+          }),
+          usage: new fields.SchemaField({
```
Forien, line 31: I'm not 100% sure yet this is needed. I just know that I had some issues about non-existing `overcast.usage` when I tried casting a spell that's not physically on Actor.

Will test more and if works without it, I will probably rever this

modules/model/item/components/overcast.js
```diff
@@ -80,11 +87,11 @@ export class OvercastItemModel extends BaseItemModel {
...
-      damage: undefined,
-      range: undefined,
-      duration: undefined,
-      target: undefined,
-      other: undefined,
+      damage: null,
+      range: null,
+      duration: null,
+      target: null,
+      other: null,
```
Forien, line 94: Because it's in schema now, it can't be `undefined`, so I changed to `null`. Doesn't impact anything.

Anyway, as above I'm not 100% sure yet this is needed and might revert this if not needed.

wfrp4e.js
```diff
@@ -77,7 +77,7 @@ import SocketHandlers from "./modules/system/socket-handlers.js";
 /*  Foundry VTT Initialization                  */
 /* -------------------------------------------- */
 
-Hooks.once("init", async function () {
+Hooks.once("init", function () {
```
Forien, line 80: Should have no impact on system itself, since nothing in this callback used `await`.

But I had some issues accessing some things defined here in `init` Hook of my module.

modules/model/item/components/physical.js
```diff
@@ -68,6 +68,22 @@ export class PhysicalItemModel extends BaseItemModel
...
+     */
+    async reduceQuantity(amount = 1) {
+        const newQuantity = Math.max(0, this.quantity.value - amount);
+
+        if (newQuantity === this.quantity.value)
+            return false;
+
+        return await this.parent.update({"system.quantity.value": newQuantity})
+    }
+
```
moo-man, line 86: It's worth noting that since the DataModel defines the minimum to be 0, you could simply update it with `system.quantity.value - amount` and it should handle it fine, I'm not sure if that's "recommended" though.

Forien: Right, it would handle it. The only reason I did that was to not trigger update if item is already at 0 quantity, but now I tested it and it handles that too 😅

PR #1876, opened 2024-01-20
Add support for RAW advancement cost option (depending on advances own)

Forien:
Per CRB, page 47:

image  
image

So RAW advancement cost for Skills should go:  
10, 10, 10, 10, 10, 10, 15, 15...

Instead, it now goes:  
10, 10, 10, 10, 10, 15, 15, 15...

My proposed change makes the advancement cost work just like described in the official rulebook. However, because some people probably want the "unofficial cost", I made the setting, which by default is disabled (so by default, advances cost are counted as before – based on number of the advance being bought)

This is also supported by the table that shows first entry as `0 to 5`, as in "if you have between 0 and 5 advancements, next one costs 10". 
It makes no sense to "buy 0th advancement". And the 1st entry for that table was part of Errata.

### DIFF
--- a/modules/hooks/init.js
+++ b/modules/hooks/init.js
@@ -116,6 +116,16 @@ export default function() {
       type: Number
     });
 
+    // Register advancements rule
+    game.settings.register("wfrp4e", "rawAdvancements", {
+      name: "SETTINGS.UseRawAdvancements",
+      hint: "SETTINGS.UseRawAdvancementsHint",
+      scope: "world",
+      config: true,
+      default: false,
+      type: Boolean
+    });
+
     // Register Fast SL rule
     game.settings.register("wfrp4e", "fastSL", {
       name: "SETTINGS.FastSL",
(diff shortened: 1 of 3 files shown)


moo-man:
This is not RAW, this is one interpretation of RAW, which is ambiguous. I am not keen on this as "RAW" rules should be enabled by default, users will question why this is not enabled and what the system actually uses. Additionally, this does not handle re-evaluating XP costs for existing characters. 

In general, I am tired of this debate, and am content with saying that the system has chosen its interpretation of the XP costs and that is simply how it's handling it.

Forien:
> This is not RAW, this is one interpretation of RAW, which is ambiguous. I am not keen on this as "RAW" rules should be enabled by default, users will question why this is not enabled and what the system actually uses. Additionally, this does not handle re-evaluating XP costs for existing characters.
> 
> In general, I am tired of this debate, and am content with saying that the system has chosen its interpretation of the XP costs and that is simply how it's handling it.

As I said this would be **disabled by default** and only be there as an option.

moo-man:
Right, and it is odd that "RAW" rules are disabled by default

PR #1886, opened 2024-02-01
auras and area template tweaks

silentmark:
Hi, 

this is my proposal to switch auras from dummy templates to hidden, persistent templates. This will allow integration with Automated Animations, as well as with potentially other modules (Levels). There are also 2 scenarios, when effects/templates are not correctly handled. If you have a token on scene, and you add an item with Aura effect - then aura will not show until any token on the scene moves. Same when you delete an item - i wasn't sure if i want to put this part of code in delete / create hooks.

In addition, there are three methods for checking the templates intersection with tokens, it is based on https://github.com/operation404/boneyard-template-tools (if you would consider using it, you might mention this module).

And this is how it looks like.
[sample.webm](https://github.com/moo-man/WFRP4e-FoundryVTT/assets/7930626/b3e2f5ba-efbc-43f3-946c-df96fa7b4cd0)


moo-man:
I've implemented my preferred method of handling aura templates in b10049a9f473504d7169458a169da6fa0a2d4b11

Please review it to consider how to include your desired features

modules/actor/actor-wfrp4e.js
```diff
@@ -105,7 +105,7 @@ export default class ActorWfrp4e extends WFRP4eDocumentMixin(Actor)
     let effects;
     if (collection == "items")
     {
-      effects = documents.reduce((effects, item) => effects.concat(item), []);
+      effects = documents.reduce((effects, item) => effects.concat(item.effects.map(x=>x)), []);
```
moo-man, line 108: `effects.map(x=>x)` does nothing. Did you mean filter?

modules/hooks/activeEffect.js
```diff
@@ -0,0 +1,13 @@
...
+    Hooks.on("createActiveEffect", async (effect, data, user) => {
+        if(game.user.isUniqueGM && effect.flags?.wfrp4e?.applicationData?.type) {
+            if (effect.flags.wfrp4e.applicationData.type == "aura" && (effect.flags.wfrp4e.applicationData.targetedAura == "target" || effect.flags.wfrp4e.applicationData.targetedAura == "all") || 
+                effect.flags.wfrp4e.applicationData.type == "area" &&
+                game.canvas?.scene) {
+                AreaHelpers.checkAreas(game.canvas.scene);
+            }
+        }
+    });
+}
```
moo-man, line 13: For one thing, where possible, I am avoiding the use of hooks and trying to consolidate it within the CRUD functions of the document itself. 

Secondly, this should already be handled by `_onCreateDescendantDocuments` in the actor.

modules/hooks/init.js
```diff
@@ -667,6 +667,19 @@ export default function() {
...
+      scope: 'world',
+      config: true,
+      type: String,
+      choices: {
+        "centerPoint": "SETTINGS.templateCollisionCenterPoint",
+        "grid": "SETTINGS.templateCollisionGrid",
+        "area": "SETTINGS.templateCollisionArea"
+      },
+  });
+
```
moo-man, line 682: My gut reaction here is that it's a bit much, the system provides the basic functionality of area effects, then modules can provide more specific alternatives if desired.

modules/hooks/templates.js
```diff
@@ -4,34 +4,34 @@ export default function()
 {
     Hooks.on("updateMeasuredTemplate", (template, data, options, user) => 
     {
-        if (game.user.id == user && (data.x || data.y))
+        if (game.user.isUniqueGM && (data.x || data.y) && !data.flags?.wfrp4e?.preventRecursive)
```
moo-man, line 7: Where is the recursion coming from here? I don't even see `preventRecursive` in this PR

modules/system/chat-wfrp4e.js
```diff
@@ -113,12 +113,44 @@ export default class ChatWFRP {
...
+                }
+              }
+              let applyData =  { effectData: [effectData] };
+              applyData.messageId = messageId;
+              target.applyEffect(applyData);
+            }
+          }
+      } else {
+        AOETemplate.fromString(event.currentTarget.text, actorId, itemId, messageId, type=="diameter").drawPreview(event);
+      }
```
moo-man, line 153: I don't understand why this has all been dumped in here, there will need to be an alternative for whatever objective is trying to be accomplished here.

modules/system/effect-wfrp4e.js
```diff
@@ -691,8 +717,7 @@ export default class EffectWfrp4e extends ActiveEffect
             areaType : "sustained", // Area - "instantaneous" or "sustained"
             renderAura : true, // Whether or not to render the measured template
 
-            targetedAura : false, // Aura - if the aura should be applied to a target and not self
-
+            targetedAura : "target",// "self", "target", "all" // Aura - if the aura should be applied to a target and not self
```
moo-man, line 720: I don't really have a mind to go through the effects again and select the correct value here. This should work similarly to normal targeted effects:

`targetedAura` is a boolean, if true, and the spell/miracle designates self/you as the target, it applies to yourself automatically. Such as, say, Aetheric Arms/Armour.

`targetedAura` might be worth renaming as well, maybe `appliedAura`.

modules/actor/actor-wfrp4e.js
silentmark: when i was testing, item.effects in concat was not treated as an array/iterable, instead it was applying whole objects collection which was breaking stuff.

modules/system/effect-wfrp4e.js
silentmark: there is a big chance, that i dont understand something related to auras. I believe there are few kinds of auras. In terms of who should be affected:
- everyone in range except token "owning" aura, 
- everyone in range including token "owning" aura - and this is the case, that i'm not sure who to achieve.

and in terms of how to apply aura: 
- whenever an item is 'invoked' - spells/miracles, maybe some other manual effects (didnt find a usecase yet (potion?))
- whenever an item is "owned" - traits, equipped weapones. 

for first case (spells), i assume that ```applicationData.documentType``` should be ```Item``` and for second (traits) it should be ```Actor```?

modules/system/area-helpers.js
```diff
@@ -1,25 +1,24 @@
...
      */
-    static isInTemplate(point, template)
-    {
-        if (template.document.t == "rect")
-        {
-            return this._isInRect(point, template);
+    static isInTemplate(token, template) {
+        let point = token.object.center;
+        //TODO: check if boneyard module is present and use it instead.
+        if (template.t == "rect") {
```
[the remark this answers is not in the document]
silentmark: yes, the goal was to handle various collision methods, and for that, we need object, probably it can be made more generic, to pass not only token but any placeable object or shape. Otherwise there would have to be another overwride method for point only.

modules/hooks/init.js
silentmark: that would be tricky to handle at this point, unless some methods of AreaHelper could be easily overwritten. If this is no-go, i can think about something else and leave only the token center implementation.

modules/hooks/templates.js
silentmark: removed, it was something old from my first appraoch to area effects.

modules/system/chat-wfrp4e.js
silentmark: i've removed it for now. the goal was to automatically apply aura to caster.

modules/hooks/activeEffect.js
silentmark: removed

modules/actor/actor-wfrp4e.js
moo-man: Then you just want `item.effects.contents`, which is an array, `item.effects` is a collection so doesn't natively concat

PR #1897, opened 2024-03-17
Effect refactor fixes

silentmark:
- various fixes, i.e. size modifires, area effect radius, weapon post to chat, unofficial grimoire
- compatibilty with Terrain Mapper module (added check for type == 'base')
- minor formatting
- additional rules implemented for endround checks (Poisoned, Stunned, Broken).


moo-man:
There will probably be some conflicts with some of my uncommitted changes, and if you can remove some of the code formatting that would be helpful

silentmark:
In this case, i guess it will be easier if you would just check if there are any bugs you haven't spot and just abondon the PR. In future i would create other PRs with dedicated changes.

modules/apps/roll-dialog/cast-dialog.js
```diff
@@ -31,6 +31,7 @@ export default class CastDialog extends SkillDialog {
 
         data.skill = spell.skillToUse;
         data.characteristic = data.skill?.system?.characteristic?.key || "int";
+        fields.unofficialGrimoire = game.settings.get("wfrp4e", "unofficialgrimoire");
```
moo-man, line 34: Fields are generally reserved for properties that are actually part of the dialog, visible to the user. This should go in `data`

modules/model/actor/standard.js
```diff
@@ -105,18 +110,18 @@ export class StandardActorModel extends BaseActorModel {
         this.computeItems();
         super.computeDerived(items, flags);
         this.runScripts("computeCharacteristics", this.parent);
+        this.computeAdvantage();
+        this.computeMove();
+        this.computeSize();
```
moo-man, line 115: Why have these been moved?

modules/system/aoe.js
```diff
@@ -68,7 +68,7 @@ export default class AbilityTemplate extends MeasuredTemplate {
     let effect = fromUuidSync(effectUuid);
     // Sometimes, the radius needs to reference the test (usually overcasting)
     setProperty(effect, "flags.wfrp4e.sourceTest",  game.messages.get(messageId)?.getTest());
-    radius = radius || effect.radius; 
+    radius = (radius || effect.radius) / 2;
```
moo-man, line 72: This seems incorrect, as it's not checking for spells, also this has already been handled in https://github.com/moo-man/WFRP4e-FoundryVTT/commit/0af72d54b768699652accbac72aa4e8d5d9071ff,

modules/system/config-wfrp4e.js
```diff
@@ -1174,13 +1174,14 @@ WFRP4E.PrepareSystemItems = function() {
...
                             await test.roll();
-                            await this.actor.applyFear(terror, name)
                             if (test.failed)
                             {
                                 if (test.result.SL < 0)
                                     terror += Math.abs(test.result.SL)
 
                                 await this.actor.addCondition("broken", terror)
+                            } else {
+                                await this.actor.applyFear(terror, name)
```
moo-man, line 1184: This is incorrect, fear should always be applied.

modules/system/config-wfrp4e.js
```diff
@@ -1726,6 +1728,18 @@ WFRP4E.PrepareSystemItems = function() {
...
+                            await test.roll();
+                            if (test.result.outcome == "success")
+                            {
+                                await actor.removeCondition("poisoned", Math.min(test.result.SL, conditionValue));
+                                msg += "<br/>Number of removed Poisoned Conditions: " + Math.min(test.result.SL, conditionValue);
+                            }
+                            else
+                            {
+                                msg += "<br/>Failed to remove Poisoned Condition";
+                            }
```
moo-man, line 1741: `appendTitle` should probably use `game.i18n.localize("WFRP4E.ConditionName.Poisoned")`

You can also use `test.succeeded` and `test.failed` getters


Also, I'm not sure about this in general because I think the handling of condition removal deserves more attention, at the very least it should be decoupled from applying the damage

modules/system/config-wfrp4e.js
```diff
@@ -1833,8 +1847,51 @@ WFRP4E.PrepareSystemItems = function() {
...
+                                let messageData = game.wfrp4e.utility.chatDataSetup(msg);
+                                messageData.speaker = {alias: this.actor.prototypeToken.name}
+                                messageData.flavor = this.effect.name
+                                return messageData
+                            }
+                            else
+                            {
+                                return this.script.scriptMessage(msg)
+                            }
+                            `
```
moo-man, line 1893: As said above, I think condition removal should wait, also there is unlocalized strings here.

modules/system/rolls/cast-test.js
```diff
@@ -190,7 +192,7 @@ export default class CastTest extends TestWFRP {
 
     // TODO handle all tooltips (when they are added) in one place
     // TODO Fix weird formatting in tooltips (indenting)
-    this.result.tooltips.miscast = this.result.tooltips.miscast.join("\n")
+    this.result.tooltips.miscast = "<ul style='text-align: left'>" + this.result.tooltips.miscast.map(t => `<li>${t}</li>`).join("") + "</ul>";
```
moo-man, line 195: This may not be needed, I have not investigated it but I believe there is a `data-tooltip-class` property we can use to style foundry tooltips

static/templates/items/item-spell-sheet.hbs
```diff
@@ -54,14 +54,14 @@
         </div>
       </div>
```
moo-man, line 56: Please revert the formatting changes here

modules/actor/actor-wfrp4e.js
```diff
@@ -952,31 +950,32 @@ export default class ActorWfrp4e extends WFRP4eDocumentMixin(Actor)
...
+        for (let data of effectData) 
+        {
+            let owningUser = game.wfrp4e.utility.getActiveDocumentOwner(this);
+            if (owningUser?.id == game.user.id)
             {
                 await ActiveEffect.implementation.create(data, {parent: this, message : messageId});
+            }   
+            else 
+            {
+                await game.wfrp4e.socket.executeOnOwnerAndWait(this, "applyEffect", {effectUuids : [], effectData : [data], actorUuid : this.uuid, messageId});
```
moo-man, line 975: What is the point of this change? Reevaluating `owningUser` multiple times is pointless as it won't change.

moo-man (commented):
I'm not very enthused by adding code to handle specific modules, I don't really like adding "base" checks. Is there any alternative solution?

modules/model/actor/standard.js
silentmark: most important is to move ComputeSize before checkWounds, otherwise woundds calculation doesn't take into account Size.

modules/system/config-wfrp4e.js
silentmark: that's interesting, in Polish book, it says that the fear should only be applied after passing the Terror test.

modules/system/config-wfrp4e.js
silentmark: ok, i thought that it might follow the same pattern as other scripts, i.e. bleeding, but maybe i will come up with something else, to remember about the tests to remove conditions.

modules/system/rolls/cast-test.js
silentmark: i guess we can remove the style, but i assume that to make it look pretty, <ul> structure has to be preserved?

static/templates/items/item-spell-sheet.hbs
silentmark: ok,  but i added it because i think there is something wrong with closure of some nodes, so if you could double-check this file.

modules/actor/actor-wfrp4e.js
silentmark: i guess i can move it to top. The goal of this change is to normalize how effect data is being passed, so that we could await the sockets always. 
I was also thinking about having a flag in ```getActiveDocumentOwner(this, gmOnly);``` and another flag 'Run as GM' in effect configuration. This way i could define an effect that should be run only as GM, but since i dont see a valid point for that yet - i only wanted to prepare this function and make it fully awaitable.

moo-man: Unless there is some pressing need I am unaware of, I don't think this change is needed. The only thing I can think of is from within a script, but that can also simply use `await game.wfrp4e.socket.executeOnOwnerAndWait(this, "applyEffect")` itself

modules/system/config-wfrp4e.js
moo-man: I think a separate script (manual trigger) for removal might work, I simply don't want to bind the effect and the removal process together.

silentmark: so, i added separaton for removal and applying damage. I like to have this reminder at end of the round, because during my games i often forget to check for players/enemies to remove conditions

modules/apps/roll-dialog/roll-dialog.js
```diff
@@ -36,7 +36,7 @@ export default class RollDialog extends Application {
         this.tooltips = new DialogTooltips();
 
         this.initialFields = mergeObject(this._defaultFields(), fields);
-        this.fields = this._defaultFields();
+        this.fields = mergeObject(this._defaultFields(), fields);
```
moo-man, line 39: What is the purpose of this?

modules/system/effect-wfrp4e.js
```diff
@@ -585,9 +585,13 @@ export default class EffectWfrp4e extends ActiveEffect
         {
             return test;
         }
+        else if (test?.data)
+        {
+            return TestWFRP.recreate(test.data);
+        }
```
moo-man, line 591: This is risky and I don't think it will work. Scripts should expect `sourceTest` to be serialized test data. If you know of any scripts that don't assume that (like trying to use getters) let me know.

silentmark: Why this is risky? I noticed one effect that was inconsistent. When run without socket it was getting testobject and worked fine, and on socket it was only getting testdata. I cannot be 100% if this wasn't custom effect, but for sure inconsistency was there

moo-man: It's risky because if tests are created during world load they may fail due to referencing documents that aren't ready yet. 

I've changed all instances of `this.effect.sourceTest.data.result` to `this.effect.sourceTest.result` if that's what you mean by inconsistency.

PR #1901, opened 2024-03-25
Extendable Area of Effect for Miracles

### LINKED ISSUE #1190: Miracles that can expand AoE
Since the last update Miracles can't use overcasts to increase AoE, which is correct. The AoE shows in the card so the player can create the template, which is great.

But a few Miracles have expanding AoE as extra uses of overcasts in their descriptions. Here is the list:

Morr:
-Destroy Undead (+Fellowship Bonus yards).
Rhya:
-Rhya's Children (+Fellowship yards).
Sigmar:
-Heed Not the Witch (+Fellowship Bonus yards).
-Soulfire (+Fellowship Bonus yards).
Taal:
-Tanglefoot (+Fellowship Bonus yards).

Looked at the Miracles in Power Behind the Throne Companion, but none has this.

Txus5012:
This PR adds a new option in the prayer sheet. When selecting "AoE" there will be an option to allow extending its area of effect.
Closes #1190 

Extendable AoE

Reading carefully at these miracles, they all extend their Area of Effect like spells, so don't need extra changes.
Also looked at the miracles in Archives of the Empire 3 and none has this.

Of course, these miracles must be updated in the compedium.


modules/model/item/prayer.js
```diff
@@ -20,6 +20,7 @@ export class PrayerModel extends OvercastItemModel
         schema.target = new fields.SchemaField({
             value : new fields.StringField(),
             aoe : new fields.BooleanField(),
+            extendableaoe : new fields.BooleanField(),
```
moo-man, line 23: Let's use camel case here, so either `extendable`, `extendableAoe` or `extendableAoE`

PR #1906, opened 2024-04-02
Conditions localization and updates

Txus5012:
- Fixed issue with stunned and fatigued at the same time by making the labels different. Applied to Poisoned too because it's the same penalty.
Dialog
- Added/Restored localization to condition dialog labels.
- Added an effect to Engaged: The actor can't make ranged attacks unless the weapon has Pistol quality.


moo-man:
I believe this falls outside of how I'm willing to localize effects, as generally the scripts are redefined by the module if the localization keys needed aren't present

Txus5012:
Just localized it because it was how it used to be. No problem.

Txus5012:
Reverted localizations. Also added an activateScript to Poisoned.

modules/system/config-wfrp4e.js
```diff
@@ -2085,7 +2089,21 @@ WFRP4E.PrepareSystemItems = function() {
...
+                    scriptData: [
+                        {
+                            trigger: "dialog",
+                            label : "@effect.name",
+                            script : `args.abort = true
+                            ui.notifications.error(game.i18n.localize("EFFECT.ShooterEngagedError"))`,
+                            options : {
+                                dialog : {
+                                    hideScript : "return !args.weapon || args.weapon.attackType == `melee` || args.weapon.properties.qualities.pistol",
+                                    activateScript : "return args.weapon.attackType == `ranged` && !args.weapon.properties.qualities.pistol"
```
moo-man, line 2102: Looks good, only thing is that I'm trying to use `weapon.isMelee/isRanged` instead of checking `attackType` directly

PR #1908, opened 2024-04-05
Localized Gender for Name Generator

Txus5012:
This allows the name generator to use localized genders. "male" and "female" can still be used.
Tested in the character creator and the name button of the NPC and Creature sheets.


Txus5012:
Added a little change. Is this a better practice? If not I can revert it.

moo-man:
I do like it better

PR #1910, opened 2024-04-08
Lingering update

Txus5012:
Lingering symptom had a few issues:
- Difficulty wasn't set in roll dialog.
- Using the difficulty keys made "Very Easy" and "Very Hard" not to work. Also, localized difficulties didn't work.
- This didn't work at all with localized symptoms.


moo-man:
I'm not going to have property keys be in the localization JSON, so an alternative way of doing this will need to be found. If the symptoms are overridden with their localized values, you should be able to use `findKey` in the difficulty config object to get the difficulty property key.

Txus5012:
I'm not sure if I'm understanding this well. I'm not localizing the difficulty keys. I just added some strings for the difficulty names without the modifier (e.g. "Easy" instead of "Easy (+40)" ).
Anyway, I see what you mean with findKey. I'll take a look at that.

Txus5012:
Added difficultyNames to config and used it for finding the difficulty key.
Also added a line break in the message, I think it looks better.

moo-man:
I was referring to this, which yeah looking at it now wasn't exactly keys, but I still think the localization JSON should strictly be for display. Anyway, it's looking better now. 

image

Txus5012:
The lang file has the names of skills, talents and traits. This is the same.

moo-man:
I've moved the PR to the master branch as I've merged `effect-refactor`

The conflict is only from `skipTargets` added to the test, so should be easy to resolve.

PR #1912, opened 2024-04-10
AoE Fix

Txus5012:
This is a fix for the AoE button in the overcast section of Spells/Prayers rolls.

The system always applied "diameter" in this AoE, now it will only apply diameter for spells.

### DIFF
--- a/modules/system/aoe.js
+++ b/modules/system/aoe.js
@@ -36,7 +36,11 @@ export default class AbilityTemplate extends MeasuredTemplate {
     {
       if (aoeString.toLowerCase().includes(game.i18n.localize("AoE").toLowerCase()))
         aoeString = aoeString.substring(aoeString.indexOf("(")+1, aoeString.length-1)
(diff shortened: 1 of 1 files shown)


modules/system/aoe.js
```diff
@@ -36,7 +36,11 @@ export default class AbilityTemplate extends MeasuredTemplate {
     {
       if (aoeString.toLowerCase().includes(game.i18n.localize("AoE").toLowerCase()))
         aoeString = aoeString.substring(aoeString.indexOf("(")+1, aoeString.length-1)
-      
+
+      let test = game.messages.get(messageId).getTest();
```
moo-man, line 40: This is clunky, the point of the diameter parameter is to be able to configure this behavior. The solution here should be to call the function with the correct diameter argument

Txus5012: No clue about this, so you can close this one if you want.

PR #1980, opened 2024-05-01
Better localization for promptLocation

Txus5012:
As it is now, promptLocation works very well in english but not in other languages.
- Title and hint aren't localized.
- Button labels aren't shown correctly. In some languages location and side are said in the opposite order. Also, the location can be male or female and the word for the side changes, and different languages can treat this differently. Because of this, I think it's better to make strings for each location.

It is also important that even if location value is translated, location key shouldn't be.


moo-man:
I didn't include this because, well, I just don't like it and would rather find an alternative, perhaps even move it to a specific function that localization modules can override.

Forien:
Maybe moving these out to `game.wfrp4e.config`?

Txus5012:
Maybe the scripts of the items could send this info?

Txus5012:
Better like this? This solves all the issues execpt the gender one, but that's a minor one.

moo-man:
It is better, yeah, though I'm considering whether or not translation modules should just overwrite the function entirely so they have complete control over how it's handled 🤷‍♀️

Issue #688, opened 2021-07-09
[Improvement] Prone at 0 Wounds


Havoclivekiller:
Per the rules, when a creature falls to 0 Wounds, they gain the Prone condition. We could have either a pop-up or a chat box (like the Apply Bleeding one).

moo-man:
I greatly dislike this rule which is why I've never gone forward to implement it, letting the GM decide to do it or not, but it could easily be a toggleable option

PR #1988, opened 2024-05-02
When a creature falls to 0 Wounds, they gain the Prone condition.

Lecht42:
Topic:
https://github.com/moo-man/WFRP4e-FoundryVTT/issues/688

### DIFF
--- a/modules/actor/actor-wfrp4e.js
+++ b/modules/actor/actor-wfrp4e.js
@@ -70,7 +70,15 @@ export default class ActorWfrp4e extends WFRP4eDocumentMixin(Actor)
     await super._onUpdate(data, options, user);
 
     if (options.deltaWounds) {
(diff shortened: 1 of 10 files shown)


Forien:
1. This is not true with Up in Arms rules, which **completely replace** the CRB rules on `Wounds, Critical Wounds, and Death` from pages 172-178
2. Would you be kind and explain the **REST** of the changes in this PR? Like removing autocalc and changes to chargen?

Lecht42:
> 1. This is not true with Up in Arms rules, which **completely replace** the CRB rules on `Wounds, Critical Wounds, and Death` from pages 172-178, so **automatically always giving out Prone** is just bad. 

Okay I'll add choise in dialog for the player.

> 3. Would you be kind and explain the **REST** of the changes in this PR? Like removing autocalc and changes to chargen?

That's not my changes, it's moo man's.

Txus5012:
Maybe this could better be a setting.

PR #2005, opened 2024-05-06
localize issue

Yasnen:


moo-man:
As vehicles are currently being worked on, I've switched the branch to `develop` so the conflicts will need to be resolved before merging.

PR #2014, opened 2024-05-08
Condition localize

Yasnen:


moo-man:
Localization of scripts and their labels should be handled by the translation modules themselves, redefining these properties as they need.

Yasnen:
The function "PrepareSystemItems" that defines statusEffects is hooked with "ready", can you generate some hooks after the process is completed or can you handle it with "init"?

moo-man:
@Yasnen Other translation devs have indicated they use the `setup` hook to replace these with the translated versions

PR #2035, opened 2024-05-16
breakdown tooltip localize issues

Yasnen:


modules/apps/roll-dialog/roll-dialog.js
```diff
@@ -188,7 +188,7 @@ export default class RollDialog extends Application {
                 this.fields[key] += this.userEntry[key];
             }
         }
-        this.tooltips.finish(this, "User Entry")
+        this.tooltips.finish(this, game.i18n.localize("BREAKADOWN.UserEntry"));
```
moo-man, line 191: I've made a typo here, if you can include a fix for that as well.

PR #2036, opened 2024-05-16
corruption strength localize issue

Yasnen:
Since Japanese translation uses a different translation for the strength of Corruption, three keys were added to the language file.

### DIFF
--- a/modules/system/chat-wfrp4e.js
+++ b/modules/system/chat-wfrp4e.js
@@ -392,7 +392,7 @@ export default class ChatWFRP {
 
   static _onCorruptButtonClicked(event) {
     let strength = $(event.currentTarget).attr("data-strength").toLowerCase();
-    if (strength != "moderate" && strength != "minor" && strength != "major")
+    if (strength != game.i18n.localize("CORRUPTION.Moderate").toLowerCase() && strength != game.i18n.localize("CORRUPTION.Minor").toLowerCase() && strength != game.i18n.localize("CORRUPTION.Major").toLowerCase())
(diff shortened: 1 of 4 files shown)


modules/system/chat-wfrp4e.js
```diff
@@ -392,7 +392,7 @@ export default class ChatWFRP {
 
   static _onCorruptButtonClicked(event) {
     let strength = $(event.currentTarget).attr("data-strength").toLowerCase();
-    if (strength != "moderate" && strength != "minor" && strength != "major")
```
moo-man, line 395: I'm not sure this will work, these are meant to be data keys, not human readable values.

scripts/xS2su09zcza9du09.js
```diff
@@ -1,8 +1,8 @@
-if (["Minor", "Moderate", "Major"].includes(this.item.system.specification.value))
+if ([game.i18n.localize("CORRUPTION.Minor"), game.i18n.localize("CORRUPTION.Moderate"), game.i18n.localize("CORRUPTION.Major")].includes(this.item.system.specification.value))
 {
 	return
 }
 
-let choice = await ItemDialog.create(ItemDialog.objectToArray({minor : "Minor", moderate : "Moderate", major : "Major"}, this.item.img), 1, "Choose Corruption Severity");
```
moo-man, line 6: You might as well include a translation for the "Choose Corruption Severity" text also

Yasnen: I don't know how to add this request because I don't know how to use git very well, and I would like to submit another pull request when I have more time because I have other script text that I would like to have as a language file.

PR #2038, opened 2024-05-16
Allow talent replacement and predefined career table

Forien:
Two changes suggestions in order to make Sea of Claws Chargen work better.

1. Allows all three Norse subspecies to share a rollable table, as defined in their config
2. Allows species and subspecies to have optional Talent replacement 

**SoC, page 56 (bottom left):**
> A Skaeling Character who generates the Pure Soul Talent as a Random Talent may take the Mark of Khorne instead.

### DIFF
--- a/modules/apps/chargen/career.js
+++ b/modules/apps/chargen/career.js
@@ -125,8 +125,11 @@ export class CareerStage extends ChargenStage {
   async addCareerChoice(number = 1) {
     let rollSpecies = this.data.species;
 
+    // If subspecies has defined specific table, and it's found, use that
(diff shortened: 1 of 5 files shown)


Txus5012:
Nice. Can the replaced talent be "any"?

UiA, p.55

> Characters from Luccini may at their option replace one of their starting Talents with Doomed.

Forien:
> Nice. Can the replaced talent be "any"?
> 
> UiA, p.55
> 
> > Characters from Luccini may at their option replace one of their starting Talents with Doomed.

I completely forgot this rule exists, so no, it's not implemented. And since my solution is a popup prompt `do you want to replace X with Y?` it could be silly here. 
Perhaps other solution is in place for that 🤔

Txus5012:
Maybe a button similar to the one that shows in repeated talents? The prompt could still work as a safety measure.

PR #2044, opened 2024-05-18
Automatic Engagement

Txus5012:
Adds a setting to allow the automatic addition of the Engaged effect after a melee attack (weapon or trait). It is added after resolving the opposed test.
Disabled by default.

Automatic Engagement


silentmark:
@Txus5012 don't you have problems with lack of permissions for actors if one of them is not GM? 

I have my own implementation, with additional remove condition on move https://github.com/silentmark/wfrp4e-pl-addons/blob/main/scripts/auto-engaged.mjs

Txus5012:
I tried it as a player and no problem.

I wouldn't remove it on move, because they can still move around their opponent (or at least many GMs understand so).

silentmark:
This part is being handled

Txus5012:
To avoid the lack of permissions issue, I made the attacker add it's own engaged in the attack test (if there is any target) and the defender add it's own engaged in the opposed test. This makes the process look a bit different.

Automatic Engagement

PR #2059, opened 2024-05-28
Generic aspect usage and sheet trigger

Forien:
This PR allows aspects to be triggered by the system, while letting modules define the implementation. 

On the same note, I am unsure if the following schema even belongs to the Generic Aspect Model, since the model itself is abstract and the children should define their own schema (and may not even use this one)

```js
schema.use = new fields.SchemaField({
  formula : new fields.StringField({initial : ""}),
  skill : new fields.StringField({initial : ""})
})
```

### DIFF
--- a/modules/hooks/init.js
+++ b/modules/hooks/init.js
@@ -774,5 +774,10 @@ export default function() {
 
     // Keep a list of actors that need to prepareData after 'ready' (generally those that rely on other actor data - passengers/mounts)
     game.wfrp4e.postReadyPrepare = [];
+
+
+    loadTemplates({
+      aspectDetails: 'systems/wfrp4e/templates/items/partials/item-aspect-details.hbs'
+    });
   });
 }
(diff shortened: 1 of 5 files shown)


Forien:
Removed the schema from `generic.js` in favour of letting children define it

Added support for dynamic partial defined by children, so they don't need to define entire Item Sheet template.

modules/hooks/init.js
```diff
@@ -774,5 +774,10 @@ export default function() {
 
     // Keep a list of actors that need to prepareData after 'ready' (generally those that rely on other actor data - passengers/mounts)
     game.wfrp4e.postReadyPrepare = [];
+
+
+    loadTemplates({
+      aspectDetails: 'systems/wfrp4e/templates/items/partials/item-aspect-details.hbs'
+    });
```
moo-man, line 781: I don't see a reason to have this separate from the rest?

Forien: Yes, you can pass an object and then call them like this:

```hbs
{{> aspectDetails}}
```
which I find easier to manage, especially in case you want to change paths to templates ;) 

However I used "dynamic partials", which can be done two ways:

1. `getPartial` is a function
```hbs
{{> (getPartial)}}
```
2. `somePartial` is a field on object
```hbs
{{> (lookup . 'somePartial')}}
```

PR #2065, opened 2024-06-03
EquippableItemModel

Forien:
# Purpose of this Pull Request
Main purpose oif this PR is to add new layer of Data Model called `EquippableItemModel`, which is a subtype of `PhysicalItemModel` representing any and all physical items that can be `equipped`. This should encompass everything, from accessories, through armour and clothing, to weapons etc. providing a standardized framework of handling those items, their state and operations on them (such as toggling equipped state, or checking if they are equipped).

This is important step to unify all variations like `item.system.equipped`, `item.system.equipped.value`, `item.system.worn` and ``item.system.worn.value` into one, cohesive system.

## What this Pull Request will allow
The "end game" of this change is to allow 3rd party (or even official) modules to easily expand on inventory and equipment system. This change follows my future idea of dynamically rendered Inventory based on Data Models, instead of hardcoded categories, and this is a step in achieving this. 

This change would allow system to simply check the Data Model and it would know if the item can be equipped (so it can show the toggle) and what to call when the toggle is clicked. Think of it as "Interface". 

## Phase 1 (finished)
1. Creation of `EquippableItemModel` Data Model, which works both as Interface and main implementation of "equippable logic"
2. Move `Armour`, `Weapon`, `Trapping` and `Container` type items to extend on `EquippableItemModel` 
3. Migrate old data to new structure
4. Modify `ActorSheetWfrp4e#_onItemToggle()` method to work with any and all Equippable items
5. Modify `ActorWfrp4e#equipPointsUsed` getter to count all Equippable items depending on their specific handling of "equipPoints". 

## Phase 2 (to-do)
1. Test system extensively to make sure no piece of code references the old getters (**especially** in case of weapons!)
2. Update all Scripts that may try to directly update items with old data structure.


Forien:
Decided that at this point it is neither needed, nor practical, to have items of Equippable Model to be displayed on Combat Sheet, so removed that from plan.

Also changing PR to "Ready for review" while I test it on my end.

PR #2088, opened 2024-06-10
[Breaking] Update grid config in manifest to v12

Forien:
Fixed manifest deprecation warning.

image

## This is breaking change so it should **only be merged for v12-only** release.

- Source: https://github.com/foundryvtt/foundryvtt/issues/9999


andrewbeard:
If you include the previous fields (gridDistance, gridUnits) in the system.json file as well as the new ones it will silence the warning but maintain backwards compatibility. Then when you drop v11 support you can just remove the previous values. This is the method that's been recommended on the Foundry Discord and has been implemented in other systems.

PR #2129, opened 2024-06-29
Load travel data files with .json extension only

grz0:
Modify `loadTravelData()` function to load `.json` files only, preventing errors with `.json.gz` files stored for nginx static compression.


silentmark:
@grz0 , czy to na pewno właściwy projekt?

grz0:
Tak.

PR #2134, opened 2024-06-30
scripts localization

silentmark:
Hi, 

I've reviewed all scripts and wanted to propose few changes, that will minimize the need of localization in module. This is definitelly not complete list, but it's always something


moo-man:
I'll merge it regardless, but were these tested at all?

silentmark:
Not all of them, but yea, during PL translation.

PR #2174, opened 2024-08-10
V12 deprecation fix, money fix, effect localization, null-exception checks.

silentmark:
As in title, few minor checks and improvements.


silentmark:
@moo-man - PR updated to latest version

PR #2175, opened 2024-08-10
Unofficial grimoire fix 

silentmark:
There was an issue with overcast not working due to data not being pasted from dialog to test and also calculation of user entries: 

```
        for(let key in this.userEntry)
        {
            if (["string", "boolean"].includes(typeof this.userEntry[key]))
            {
                this.fields[key] = this.userEntry[key]
            }
            else if (Number.isNumeric(this.userEntry[key]))
            {
                this.fields[key] += this.userEntry[key];
            }
        }
        this.tooltips.finish(this, "User Entry")
```

since overcast value in cast dialog is a numeric field, instead of passing it to test, it was either incremented on every input change, or treated as NaN, due to null + number calculation.

Also, minor typo and translation in base dialog.


silentmark:
@moo-man - this PR has been updated to latest changes.

modules/apps/roll-dialog/cast-dialog.js
```diff
@@ -13,6 +13,13 @@ export default class CastDialog extends SkillDialog {
         return options;
     }
 
+    constructor(fields, data, resolve, options) {
+        super(fields, data, resolve, options);
+
+        this.initialFields.overchannelling = this.fields.overchannelling ?? 0;
+        this.fields.overchannelling = this.fields.overchannelling ?? 0;
+    }
```
moo-man, line 21: Is there a reason this isn't part of an overridden `_defaultFields` function? See https://github.com/moo-man/WFRP4e-FoundryVTT/blob/master/modules/apps/roll-dialog/roll-dialog.js#L184

modules/apps/roll-dialog/cast-dialog.js
```diff
@@ -69,6 +76,18 @@ export default class CastDialog extends SkillDialog {
...
 
+    async getData() 
+    {
+        let context = await super.getData();
+        if (this.userEntry.overchannelling) {
+            let overchannelling = parseInt(this.userEntry.overchannelling) ?? 0;
+            this.userEntry.overchannelling = overchannelling;
+            this.fields.overchannelling = overchannelling;
+            context.fields.overchannelling = overchannelling;
+        }
```
moo-man, line 87: Not sure what the point of this is? This should all be handled natively as any other field.

static/templates/dialog/base-dialog.hbs
```diff
@@ -27,7 +27,7 @@
       <label>{{localize "DIALOG.DialogModifiers"}}</label>
       <ul class="modifier-list">
         {{#each data.scripts}}
-            <li class="modifier {{#if this.isActive}}active {{else if this.isHidden}}hidden{{/if}}" data-tooltip="{{this.effect.name}} {{#if this.options.targeter}}(Target){{/if}}" data-tooltip-direction="RIGHT" data-index="{{@key}}" >&#8226; <a>{{this.Label}}</a></li>
+            <li class="modifier {{#if this.isActive}}active {{else if this.isHidden}}hidden{{/if}}" data-tooltip="{{this.effect.name}} {{#if this.options.targeter}}({{localize 'Target'}}){{/if}}" data-tooltip-direction="RIGHT" data-index="{{@key}}" >&#8226; <a>{{this.Label}}</a></li>
         {{/each}}
       </ul>
     </div>
```
moo-man, line 33: This should probably just use the partial provided by the warhammer library (which needs to be localized too)

static/templates/dialog/spell-dialog.hbs
```diff
@@ -4,25 +4,27 @@
...
-    </label>
+    <input class="input-text" type="text" name="overchannelling" value="{{fields.overchannelling}}" data-dtype="Number" title="{{localize 'DIALOG.OverchannellingTooltip'}}" />
   </div>
+  {{#if false}}
+    <div class="form-group">
+      <label title="{{localize 'DIALOG.Quickcasting'}}">{{localize "DIALOG.Quickcasting"}}
+        <input title="{{localize 'DIALOG.QuickcastingTooltip'}}" type="checkbox" name="quickcasting" data-dtype="Boolean" {{checked fields.quickcasting}} />
+      </label>
+    </div>
+  {{/if}}
```
moo-man, line 29: I don't really use `title` or `data-dtype` anymore.

modules/apps/roll-dialog/cast-dialog.js
silentmark: I will test it out

modules/apps/roll-dialog/cast-dialog.js
silentmark: There was a problem with Numeric fields for userEntry. Maybe it's now handled better in Library. The problem was, that overchannelling was always treated as string, and in result, onFieldChange was returning either NaN as text or undefined. I will test it out.

static/templates/dialog/base-dialog.hbs
silentmark: it seams that none of the partial templates are being used in WFRP4e system so far. as of Lib translation, I might need to prepare PR there anyway.

static/templates/dialog/spell-dialog.hbs
silentmark: copy-paste, i will remove those.

modules/apps/roll-dialog/cast-dialog.js
silentmark: resolved

modules/apps/roll-dialog/cast-dialog.js
silentmark: resolved, it works fine without this change.

silentmark:
@moo-man - went through comments and updated PR.

PR #2176, opened 2024-08-10
"other" overcast custom value, customizable overcast table per lore (i.e. Dhar)

silentmark:
This is small request that would allow two things: 
- customize overcast tables per lore, i.e. WoM doesn't describe overcast rules for Dhar, this option would allow to define custom table for Dhar spells that would have lower values. 
- customize "other" value for overcast rules instead of hardcoded "2". this would allow to customize behavior of some spells from unofficial grimoire, i.e. Ward (+3 SL to increase Ward power) or Terryfing and few others. In addition, this would go beyond, and allow to "define" a formula tha would increase cost of "other" overcast for each step in similar way to table.


silentmark:
@moo-man - PR updated to latest changes.

modules/model/item/components/overcast.js
```diff
@@ -19,6 +19,7 @@ export class OvercastItemModel extends BaseItemModel {
                additional  : new fields.StringField(),
                characteristic  : new fields.StringField(),
                bonus : new fields.BooleanField(),
+               cost : new fields.StringField()
```
Forien, line 22: I don't see this referenced anywhere else other than on the Sheet. Is this utilized anywhere? 

If so, what is the point of having this as well as "other" in config?

static/templates/items/item-spell-sheet.hbs
```diff
@@ -279,13 +279,19 @@
                 <a class="item-checkbox" data-target="system.overcast.valuePerOvercast.bonus">{{localize 'Bonus'}}<i
                     class="far fa-circle"></i></a>
               {{/if}}
+            {{/if}}
+            <div class="form-group">
+              <label class="label-text">Koszt zmiany</label>
```
Forien, line 284: Needs translation (if it is to remain, look previous comment)

modules/model/item/components/overcast.js
silentmark: The idea is to be able to create a formula, that can be calculated per overcast 'level' its in WoM cast test used - 
image

I didnt add it to regular CastTest, as there overcast rules are super straightforward.

static/templates/items/item-spell-sheet.hbs
silentmark: done

modules/system/rolls/wom-cast-test.js
```diff
@@ -79,21 +84,24 @@ export default class WomCastTest extends CastTest {
...
-      } 
-      // Other options are not in the table, so assume cost is 2 per original rules
+      }
       else if (choice == "other") {
-        if (2 > overcastData.available)
+        if (overcastData.valuePerOvercast.cost) {
+          otherCost = parseInt(eval(overcastData.valuePerOvercast.cost
+                          .replace("{{current}}", overcastData.usage[choice].current)
+                          .replace("{{count}}", overcastData.usage[choice].count)));
+        }
```
moo-man, line 101: I'm not sure I understand what's going on here?

static/templates/items/item-spell-sheet.hbs
```diff
@@ -279,12 +279,18 @@
...
               {{/if}}
+            {{/if}}
+            <div class="form-group">
+              <label class="label-text">{{localize 'SHEET.OtherCost'}}</label>
+              <div class="input-box">
+                  <input class="input-text" placeholder="{{localize "SHEET.OtherCostPlaceholder"}}" data-dtype="String" type="text"
+                    name="system.overcast.valuePerOvercast.cost"
+                    value="{{system.overcast.valuePerOvercast.cost}}" />
+              </div>
+            </div>
```
moo-man, line 290: Can you hide or otherwise denote this as being a WoM specific field?

moo-man: @silentmark I'm good to merge this once resolved here

silentmark: That's the motivation ;) I will finsh all my conflicts and reported issues today night

silentmark: I hope that    {{#if (settings "useWoMOvercast")}} is enough.

PR #2178, opened 2024-08-10
Miscellaneous changes that would make my life easier

silentmark:
Most of those changes are rather cosmetic and shouldn't affect system in any way. 
1. change the way how effects are sent via socket to users - always send full effectData instead of sometimes effect data and sometimes uuids. 
2. merge data that is comming from addCondition.
3. It seems that size modifier didnt work right, there is a proposal to fix it.
4. extract computeAmbidextrous, this would allow me to playaround with twohanded fight. 
5. minor change of groupadvantage ui update - sometimes i was getting wierd "undefined" exceptions, that were bloating the console for users. Don't know from where it was comming. 
6. sometimes radius calculation was causing problems, this is propsal to check if calculation is needed when radius is just a number.


modules/apps/roll-dialog/attack-dialog.js
```diff
@@ -83,10 +83,10 @@ export default class AttackDialog extends SkillDialog
         let sizeDiff = this.actor.sizeNum - target.actor.sizeNum
         let sizeModifier = 0
         // Attacking a larger creature with melee
-        if (sizeDiff < 0 && (this.item.attackType == "melee" || target.actor.sizeNum <= 3)) 
+        if (sizeDiff < 0 && (this.item.attackType == "melee" || target.actor.sizeNum >= 3))
```
silentmark, line 86: I think this is the right check or am i missing something?

modules/apps/roll-dialog/weapon-dialog.js
```diff
@@ -120,14 +120,18 @@ export default class WeaponDialog extends AttackDialog {
 
     if (this.item.offhand.value && !this.item.twohanded.value && !(this.item.weaponGroup.value == "parry" && this.item.properties.qualities.defensive)) 
     {
-      this.fields.modifier += -20
-      this.tooltips.addModifier(-20, game.i18n.localize("SHEET.Offhand"))
+      this.computeAmbidextrous()
```
silentmark, line 123: I would like to add this additiona method, because this way i may override it in one of my house rules.

modules/hooks/settings.js
```diff
@@ -18,14 +18,14 @@ export default function() {
                 })
             }
             // Update counter values, can't just use ui.combat because there might be popped out combat trackers
-            [ui.combat].concat(Object.values(ui.windows).filter(w => w instanceof CombatTracker)).forEach(tracker => {
+            let trackers = [ui.combat].concat(Object.values(ui.windows).filter(w => w instanceof CombatTracker));
```
silentmark, line 21: that's actually wierd - but i noticed that sometimes, group advantage updates breaks using original code - it was throwing exception at [ui.combat].concat()...().foreach(). no idea why.

modules/system/effect-wfrp4e.js
```diff
@@ -168,6 +168,7 @@ export default class ActiveEffectWFRP4e extends WarhammerActiveEffect
 
     get radius()
     {
+        if (typeof this.applicationData.radius === 'number') return this.applicationData.radius;
```
silentmark, line 171: if radius is a number - parsing it as roll was throwing an exception for me, but again - maybe i was doing somethig wrong?

modules/hooks/settings.js
```diff
@@ -7,8 +7,8 @@ export default function() {
         // Then, make sure that change is reflected in the counter on the combat tracker (if the update was made by a different user)
         if (setting.key == "wfrp4e.groupAdvantageValues")
         {
-            ui.notifications.notify(game.i18n.format("GroupAdvantageUpdated", {players : setting.value.players, enemies : setting.value.enemies}))
-
+            let message = game.i18n.format("GroupAdvantageUpdated", {players : setting.value.players, enemies : setting.value.enemies});
+            ui.notifications.notify(message)
```
moo-man, line 11: Is `message` used somewhere else? Not sure what the intention here is

modules/system/effect-wfrp4e.js
moo-man: This looks like it's from before the library integration, so it'll need to be updated.

modules/hooks/settings.js
silentmark: Its really hard for me to explain, but sometimes i saw very strange exceptions being thrown in this method - thats why i wanted to split it into atomic steps.

modules/system/effect-wfrp4e.js
silentmark: looks like it, i will remove that.

silentmark:
@moo-man - went through comments and updated PR.

silentmark:
You refer to this change related to combat tab and notifications?

moo-man:
Right, and the current sizeNum check is intentional and doesn't need to be changed.

silentmark:
Ok, i will revert those, btw so can you explain to me how this sizeNum works, cause I feel like it's not how it is intended?

silentmark:
I think all your comments have been resolved.

PR #2237, opened 2024-09-28
add `isArmour` and `isWeapon` to Item Properties

Forien:


modules/model/item/armour.js
```diff
@@ -52,6 +52,10 @@ export class ArmourModel extends PropertiesMixin(EquippableItemModel) {
     }
   }
 
+  get tags() {
+    return super.tags.add("armour");
+  }
```
moo-man, line 57: Not sure why this is needed? Should be handled via the base model https://github.com/moo-man/WFRP4e-FoundryVTT/blob/66976be71b54fd4617e4d8383d68d0d41d6712d0/modules/model/item/components/base.js#L20

Forien: Well, if the parent's type is `wfrp4e-archives3.armour` for example, then there is no `armour` tag at all.

Forien: Unless the expectation is for derived models to provide that tag? 🤔

moo-man: Oh right, yeah, ArmourV2 should be adding the `armour` tag. Each item will have a tag for its type, but it can also opt into other tags.

PR #2238, opened 2024-09-28
minor fixes to various parts of the system

Forien:
Not sure if I was tripping, but `game.system.model` was not working for me at all and `game.model` does. 

Probably best to double check if it's really broken before merging this PR


moo-man:
I think game.system.model is from pre-V11

PR #2368, opened 2025-01-29
Make tables path uniform for all miscasts across both Cast and Channel Tests

Forien:
As far as I know, miscasts do not stack, but only worst appears.

Wasn't sure about grimoire, left it with unique key.


moo-man:
We need a separate null table, as if you have a Major Miscast, and it turns into it Minor Miscast, we need to display both the crossed out major miscast and show the actual minor miscast, so the nulled versions need separate entries.

Forien:
I see. So I will change it only so channelling mimics casting

PR #2395, opened 2025-02-24
add description details to npcs and creatures

silentmark:
I've moved some of the model properties from character to standard model - for NPC and Creatures. This incudes: 
- hair colour, eyes colour, age, width, height, specials. 

Please accept this change :-) 

it will help me with: 
- support for automate token height in context of Wall-height and therefore, autocover/vision calculation 
- better support for my AI-gen npcs, which i want to get back to.

### DIFF
--- a/static/templates/sheets/actor/creature/creature-notes.hbs
+++ b/static/templates/sheets/actor/creature/creature-notes.hbs
@@ -1,17 +1,50 @@
 <section class="tab {{tab.cssClass}}" data-group="primary" data-tab="{{tab.id}}">
-    {{#if (isGM)}}
-        <div class="editor-section">
-            <div class="header">{{localize "GM Notes"}}</div>
-            {{formInput fields.details.fields.gmnotes.fields.value value=source.system.details.gmnotes.value enriched=enriched.system.details.gmnotes.value toggled=true}}
+
+    <div class="header-row">
+        <div class="header-group">
+            <input type="text" name="system.details.age.value" value="{{system.details.age.value}}">
+            <label>{{localize "Age"}}</label>
         </div>
+        <div class="header-group">
+            <input type="text" name="system.details.height.value" value="{{system.details.height.value}}">
+            <label>{{localize "CharacterHeight"}}</label>
+        </div>
+        <div class="header-group">
+            <input type="text" name="system.details.weight.value" value="{{system.details.weight.value}}">
+            <label>{{localize "Weight"}}</label>
+        </div>
+        <div class="header-group">
+            <input type="text" name="system.details.haircolour.value" value="{{system.details.haircolour.value}}">
+            <label>{{localize "Hair Colour"}}</label>
+        </div>
+        <div class="header-group">
+            <input type="text" name="system.details.eyecolour.value" value="{{system.details.eyecolour.value}}">
+            <label>{{localize "Eye Colour"}}</label>
+        </div>
+    </div>
+    <div class="header-row">
+        <div class="header-group">
+            <input type="text" name="system.details.distinguishingmark.value"
+                value="{{system.details.distinguishingmark.value}}">
+            <label>{{localize "Distinguishing Mark"}}</label>
+        </div>
+    </div>
+
+    {{#if (isGM)}}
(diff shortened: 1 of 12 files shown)


static/templates/sheets/actor/creature/creature-notes.hbs
```diff
@@ -1,17 +1,50 @@
...
+            <label>{{localize "Distinguishing Mark"}}</label>
+        </div>
+    </div>
+
+    {{#if (isGM)}}
+    <div class="editor-section">
+        <div class="header">{{localize "GM Notes"}}</div>
+        {{formInput fields.details.fields.gmnotes.fields.value value=source.system.details.gmnotes.value
+        enriched=enriched.system.details.gmnotes.value toggled=true}}
+    </div>
```
moo-man, line 38: I think I'm still going to say that these fields on the creature sheet aren't necessary.

style/sheets/actors/_npc.scss
```diff
@@ -5,6 +5,25 @@
...
+                input {
+                    background: none;
+                    border: none;
+                    line-height: 1.5rem;
+                    height: 1.5rem;
+                    padding: 0 0.1rem;
+                }
+            }
+        }
+    }
```
moo-man, line 26: If you're redefining these styles, why are they still called `.header-x`?

static/templates/sheets/actor/creature/creature-notes.hbs
silentmark: It would be awesome to have it consistent, especially that creatures are also orcs, goblins, gors etc, for which those stuff make sense. I think even Vampires are actual creature objects.

style/sheets/actors/_npc.scss
silentmark: I'm open for suggestions about better naming/structuring of the class names. I didn't want to mess up too much with the scss files, but i guess that it would be possible to "share" in smart way styles for those fields that are inside of character main tab and those ones. Any preferences?

static/templates/sheets/actor/creature/creature-notes.hbs
moo-man: It doesn't really follow the original (though not always followed) paradigm of NPCs are specific characters, while creatures are generic. 

There's always been talk about there not being much distinction between NPCs and Creatures, and I think these fields are decent as a "feature" of NPCs.

These fields are fine to stay in the standard model, but I don't see any benefit to them being on the sheet as they will be rarely, if ever, used for creatures.

style/sheets/actors/_npc.scss
moo-man: `detail-x` is probably fine

static/templates/sheets/actor/creature/creature-notes.hbs
silentmark: well. at least having them in standard model would partially help. there is always way to extend ui. if you are fine with that, i will remove the UI part changes, but anyway its :(

moo-man: Yeah go ahead and remove it, but as a consolation, I am currently considering letting NPCs and Creatures use either NPC or Creature sheet, so if someone wanted a Creature to have the NPC fields, they can simply switch to the NPC sheet. 

Also considering deprecating creatures and just having two different sheets for NPCs: NPC and Creature, not sure yet.

style/sheets/actors/_npc.scss
silentmark: done

static/templates/sheets/actor/creature/creature-notes.hbs
silentmark: done

style/sheets/actors/_npc.scss
silentmark: details-row and details-cell would ok?

moo-man: Sure, using the same as the header is also fine, so detail-group/row. Doesn't really matter they'll likely get changed when merged with the v13 changes anyway

PR #2407, opened 2025-03-17
feat: Add option to remove resetting targets on each roll

Tiamanti:
Adds a client side setting to not reset targets after each roll.
Idea is that sometimes things don't die in 1 roll and need to retargeted.
This allows GM to have it reset but allows players to turn it off as they keep hitting the same thing.

Can rewrite this to be global and/or a select kind of setting ex: Reset for All, Reset for GM, Don't reset weapon rolls, etc...

image


Tiamanti:
Reverted to draft as this needs more work.

PR #2428, opened 2025-04-05
Documentation Button

Txus5012:
This PR adds a button in the settings tab that opens a window with the wiki for WFRP 4e.

Documentation button


moo-man:
This has already been implemented in v13, though feel free to PR it to use an iframe if you feel that is worth it

PR #2430, opened 2025-04-15
Static Point Blank Range Houserule implemented

marcin-jablonski:
Added homebrew* rule for static Point Blank range.

It has been discussed widely that Point Blank range in some cases is weirdly big (especially in cases of Elf Bows), and apparently (*) there was intent from early creators that PB was supposed to be fixed at some distance (4 yds if I recall correctly). This homebrew implements that.


moo-man:
Thanks however I won't be accepting this for V12, you'll need to use the `v13` branch and implement it with the changes there.

PR #2434, opened 2025-04-21
Namespaces (Warnings Begone)

Forien:
Basically went through the `Foundry\client\client.mjs` and checked all classes between lines 195–604


src/hooks/token.js
```diff
@@ -2,15 +2,6 @@ import WFRP_Utility from "../system/utility-wfrp4e.js";
...
 export default function() {
-  // Adds tooltips to conditions in the condition menu
-  Hooks.on("renderTokenHUD", async (obj, html) => {
-    for (let condition of html.find("img.effect-control")) {
-      condition.title = game.wfrp4e.config.conditions[condition.dataset["statusId"]]
-      if (condition.dataset["statusId"] == "dead")
-        condition.title = "Dead"
-    }
-  })
-
```
Forien, line 13: This I removed because it would double the tooltips. 
image

PR #2435, opened 2025-04-22
Implementing custom Token Ruler with support for WFRP4e movement rules for Movement Actions

Forien:
There are several available actions, all defined by core Foundry: `walk`, `climb`, `crawl`, `fly`, `swim`, `blink` and `displace`;
I removed `burrow` (not relevant) and `jump` (requires Movement 6 just to get 2 yards...)
I also added:
- Color coding for Token Ruler
  - Green – Normal move (Walk/Fly/Crawl)
  - Yellow – Double move (Run)
  - Red – excess move
  - Dark red - immobile / max movement for given action is 0 or less
- Flight Trait enabling Fly action and setting it's max distance
- Swim Skill and Amphibious Trait enabling Swim action and Amphibious removing max distance penalty
- Crawl having max distance of half Movement
- Climb having max distance of half normal rate (Walk/Run)
- Some conditions will warn that the Token should be immobile (`surprised`, `entangled`, and `unconscious`)
- Other conditions will hint that the Token is restricted – it shouldn't move or its move might be penalized 
  - `prone` - should use Crawl action, not automated, 
  - `engaged` - possibly don't want to move, 
  - `stunned` - halved movement, automated

[Image from Gyazo](https://gyazo.com/53b6c7173349f3e43bd323ee25990ee3)

There is also native support for vanilla Regions that can change "costs" for moving through them.
[Image from Gyazo](https://gyazo.com/bfa26959fb69acd01f3052f4b2515d14)

## For Consideration
Potentially it should be possible and could be worth considering: extending the `ModifyMovementCostRegionBehaviorType` Region Behaviour to include a `biome` or `terrain type` as a `<select>` input, and then some Traits and Talents could "ignore terrain penalties". But I haven't included that


Forien:
Here is how hints look with multiple movement-inhibiting conditions. This "warning" is purely visual and `Immobile` takes precedence over `Restricted`

image

PR #2461, opened 2025-05-09
Attach GenericActions listener to Item, Actor and Journal Sheets

Forien:
Links such as:
```
@Credit[2ss]
@Pay[2gc]
@Fear[2]
@Table[critbody]
```
are not clickable in any Document Sheets, as those Sheets lack listeners.

This PR adds frame listeners to Item, Actor and Journal Entry Sheet classes.


moo-man:
I liked the fact that was able to remove the subclassed Journal stuff from pre-V13, can this be moved to the journal sheet hook?

Forien:
I thought subclassing would be better, but sure, this can easily be handled in render Hook. Already pushed the change

PR #2476, opened 2025-05-15
Colors for ruler segment

Txus5012:
This PR adds colors for the ruler segment. The same as the grid highlight. Specially useful for gridless maps.
Ruler
Also used the same colors for immobile. Added a blue color for blink and displace.

I noticed that in gridless maps, many times when moving exactly to the speed limits (walk and run) it shows the next color. I think that is because there are some unseen decimals.

Ideally, the cursor should snap to that limit when it's close, like drag ruler did. But I have no idea on how to do that.

### DIFF
--- a/src/canvas/token-ruler.js
+++ b/src/canvas/token-ruler.js
@@ -75,23 +75,27 @@ export default class TokenRulerWFRP extends foundry.canvas.placeables.tokens.Tok
     return foundry.utils.mergeObject(style, color);
   }
 
+  /** @override */
   _getSegmentStyle(waypoint) {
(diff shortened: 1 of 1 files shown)


Forien:
I specifically didn't want to color the ruler segments, because I like the idea of seeing who is dragging that token. 

Maybe only apply those colors when on gridless scenes? Or more settings? 😅 

I guess it's up to Moo :D

Txus5012:
I'm OK with it being only for gridless scenes.

The waypoints still have the player's color, including the initial one. So you can see who is dragging in these scenes too.

PR #2481, opened 2025-05-20
Add god to `Bless` talent's name even if there is no `(Any)`

Forien:


scripts/IupskvzvoGyD2H5o.js
```diff
@@ -31,6 +31,9 @@ if (god)
     {
         this.script.notification(`Could not find any Blessings associated with ${god}.`)
     }
-    this.item.updateSource({name : this.item.name.replace("Any", god)})
+    if (this.item.name.includes("Any"))
+        this.item.updateSource({name: this.item.name.replace("Any", god)});
+    else
+        this.item.updateSource({name: this.item.name + ` (${god})`});
```
moo-man, line 37: Small tweak here using a helper, `this.item.updateSource({name : this.item.setSpecifier(god)})` should cover both cases.

PR #2483, opened 2025-05-21
[Docs] Expanded the Advanced Topics documentation

Forien:
Expanded the Advanced Topics documentation with comprehensive guides on adding Lores, Species, Subspecies and World Scripts. Included specific code examples.

### DIFF
--- a/docs/pages/advanced/species.md
+++ b/docs/pages/advanced/species.md
@@ -21,6 +21,7 @@ Config data can be found under the `game.wfrp4e.config` as soon as `init` hook.
 - [Creating Career's Rolltable](#creating-careers-rolltable)
 - [Example 1: Add Wolfkin Species](#example-1-add-wolfkin-species)
 - [Example 2: Add "Arctic" Subspecies to the Wolfkin Species](#example-2-add-arctic-subspecies-to-the-wolfkin-species)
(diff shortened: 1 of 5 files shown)


docs/pages/advanced/species.md
```diff
@@ -2,4 +2,294 @@
...
+config.speciesCareerReplacements.wolfkin = {
+  "Flagellant": "Hunter"
+}
+```
+The above will allow Wolfkin to pick `Hunter` career in case they rolled `Flagellant`
+
+{: .highlight}
+> This is not typically used for new Species, as you should provide a dedicated Random Career Table
+> 
+> For existing Species it is also recommended to use the `game.wfrp4e.utility.mergeCareerReplacements` helper method instead of direct definitions to avoid overwriting.
```
moo-man, line 159: I feel like this is important enough to have an example

Forien: Fair, but I don't feel it's good to have in a page about _adding new Species_, perhaps this could be an example in World Scripts page?

moo-man: Why not? It's specific to adding species, isn't it just a one liner?

My highlight is a bit off, I specifically been the merge function

Forien: Added full example at the end of the file and linked it here

PR #2508, opened 2025-05-31
fixed deprecation warnings

Txus5012:
Fixes for some deprecation warnings.


Txus5012:
Reverted one of the fixes because it messed the character generator.

PR #2518, opened 2025-06-03
Add support for Gromril armour type

Jagusti:
I've not kept strict alphabetical order as Gromril is so much rarer than the standard four options.


moo-man:
I might've miscommunicated, this should be done in the module itself.

PR #2540, opened 2025-06-26
Add Critical Deflection

Tiamanti:
Adds Critical Deflection mechanic as option next to Critical table wound.
Uses existing Hack mechanic.
image
Also fixes critical roll not being colored in message.
image

### DIFF
--- a/src/documents/actor.js
+++ b/src/documents/actor.js
@@ -737,6 +737,10 @@ export default class ActorWFRP4e extends WarhammerActor
         updateMsg += `<br><a data-action="clickTable" class="action-link critical-roll" data-modifier="-20" data-table = "crit${opposedTest.result.hitloc.value}" ><i class='fas fa-list'></i> ${game.i18n.localize("Critical")} (-20)</a>`
       else
         updateMsg += `<br><a data-action="clickTable" class="action-link critical-roll" data-table = "crit${opposedTest.result.hitloc.value}" ><i class='fas fa-list'></i> ${game.i18n.localize("Critical")}</a>`
+
+      if (modifiers.ap.nonmetal > 0 || modifiers.ap.metal > 0) {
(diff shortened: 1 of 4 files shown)


src/model/message/test.js
```diff
@@ -138,4 +140,32 @@ export class WFRPTestMessageModel extends WarhammerTestMessageModel
     test.moveVortex();
   }
 
+    static async onApplyCriticalDeflection(ev)
+    {
+        const controlledTokens = [...canvas.tokens.controlled, ...game.user.targets].map(t => t.actor);
+        const targetIntersection = controlledTokens.filter(t => this.test.targets.includes(t));
+        const hitLoc = this.test.hitloc.result;
```
moo-man, line 147: I have generally tried moving away from using controlled tokens.

The paradigm I go for is user targets, falling back to targets of the test if there are no user targets.

Tiamanti: My idea with including controlled tokens here would be that player could click this if they were the target of Critical Hit instead of having to target their own token but if it's the form I'll go with it.

moo-man: That would be covered by `falling back to targets of the test` right?

Tiamanti: In changed version yes.

Tiamanti: Moved this to attack-test.js because I realised this didn't work for Trait Attacks

Tiamanti:
@moo-man I've refactored few things but also added fix to #2594 .

It's a partial fix because the same issue is present in several scripts as AOE3 armour is under `itemTypes."wfrp4e-archives3.armour"` and not `itemTypes.armour` and I think I know what those scripts do but would like a go/no go on those.

moo-man:
ItemTags.armour should cover both types of armour

moo-man:
Yeah don't see any reason to exclude it, unless it has some different rules I don't remember

PR #2549, opened 2025-07-05
Increase quantity of item on purchase

Tiamanti:
If actor has said item in their inventory then just increase the quantity instead of adding new instance.
Works with items at 0 quantity.
Performance tested with 10x slow CPU and actor with 100 items of type and the search is usually completed in 0.4ms

### DIFF
--- a/src/model/message/posted-item.js
+++ b/src/model/message/posted-item.js
@@ -83,8 +83,26 @@ export class PostedItemMessageModel extends WarhammerMessageModel {
     let paid = await game.wfrp4e.market.handlePlayerPayment({payString : game.wfrp4e.market.amountToString(this.itemData.system.price), itemData : this.itemData});
     for(let actor of paid)
     {
-      await actor.createEmbeddedDocuments("Item", [this.itemData], {fromMessage: this.parent.id})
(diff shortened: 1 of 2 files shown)


src/model/message/posted-item.js
```diff
@@ -83,8 +84,18 @@ export class PostedItemMessageModel extends WarhammerMessageModel {
...
+          && i._stats.compendiumSource === this.itemData._stats.compendiumSource);
+      if(hasItem) {
+        await actor.updateEmbeddedDocuments("Item", [{
+          _id : hasItem.id,
+          "system.quantity.value" : hasItem.system.quantity.value + this.itemData.system.quantity.value,
+        }], {fromMessage: this.parent.id});
+        ui.notifications.notify(game.i18n.format("MARKET.ItemAppended", { item: this.itemData.name, actor : actor.name, quantity: hasItem.system.quantity.value }));
+      } else {
+        await actor.createEmbeddedDocuments("Item", [this.itemData], {fromMessage: this.parent.id});
+        ui.notifications.notify(game.i18n.format("MARKET.ItemAdded", { item: this.itemData.name, actor : actor.name }));
```
moo-man, line 97: You can add format to the second argument, like

`ui.notifications.info("key", {format : {...yourArguments}});`

Issue #2553, opened 2025-07-06
[REQUEST] Put current Career Group at the top of Career Advancement dialog


Forien:
When advancing, it would save a lot of scrolling if current Carrer Group would be placed at the top, then rest Career Groups of the same Class and then Career Groups out of Class.

Currently in some Classes like Warriors or Academics, the amount of Career Groups is huge and they are sorted alphabetically, meaning that someone having career group that uses letters like P-Z will most likely have to scroll a lot, which causes players to sometimes miss their career group and waste time. 

---

Different colour would also be nice. For example keep current "Out of Class - Red", "Same Class - Blue", but maybe add "Same Career Group - Green/Yellow"?

---

Additional minor QoL: Somehow marking (bold or underline for example) current Career (or somehow displaying Tier of current Career above the list) could be potentially helpful.

PR #2555, opened 2025-07-06
Improvements to Carrer Selector dialogue

Tiamanti:
Implements #2553 
What was done:

- Moves current Carrer Group to the top of Carrer Selector
- Marks current carrer group and current carrer level
- Disables selection of current carrer level

image
image

### DIFF
--- a/src/apps/career-selector.js
+++ b/src/apps/career-selector.js
@@ -63,7 +63,10 @@ export default class CareerSelector extends  HandlebarsApplicationMixin(Applicat
         const currentCareerGroup = this.currentCareer.system.careergroup.value;
         this.careers = await warhammer.utility.findAllItems("career", game.i18n.localize("CAREER.Loading"), true, ["system.careergroup.value", "system.level.value", "system.class.value"])
         this.careers = this.careers.sort((a, b) => a.system.careergroup.value > b.system.careergroup.value ? 1 : -1)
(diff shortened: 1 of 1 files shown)


src/apps/career-selector.js
```diff
@@ -59,9 +60,11 @@ export default class CareerSelector extends  HandlebarsApplicationMixin(Applicat
         {
             return 
         }
+        const currentCareerGroup = this.currentCareer.system.careergroup.value;
         this.careers = await warhammer.utility.findAllItems("career", game.i18n.localize("CAREER.Loading"), true, ["system.careergroup.value", "system.level.value", "system.class.value"])
         this.careers = this.careers.sort((a, b) => a.system.careergroup.value > b.system.careergroup.value ? 1 : -1)
-        this.careers = this.careers.filter(i => 
+            .sort((a, b) => a.system.careergroup.value === currentCareerGroup ? -1 : b.system.careergroup.value === currentCareerGroup ? 1 : 0)
+        this.careers = this.careers.filter(i =>
```
moo-man, line 67: I make it a rule which I, to my knowledge, have only broken once in some obscure piece of code, to avoid nesting ternary operators in each other, for obvious readability reasons. 

Let's just do something more readable, like if/else

Tiamanti: I think if/else would look cumbersome but unshift version is more readable. Refactored

Issue #2566, opened 2025-07-18
Color indication roll type


Mrak7439:
The wfrp system on version 12 of Foundry had a great feature - a color indication of the type of roll in the chat. The edges of the "scroll" where the roll is displayed in the chat window were brown for the public roll. The gray edges are for GM roll and Self roll. The blind roll was displayed in purple. 

Suggestion: To return the color indication of the type of roll, as it was before the transition to v13.

<img width="367" height="543" alt="Image" src="https://github.com/user-attachments/assets/dcd7dd75-2a73-46b4-82e4-88bb23d9fe91" />

PR #2580, opened 2025-08-01
refactor(2566): Restore old message borders

Tiamanti:
Addresses #2566 
Restores old chat message colored borders.
Differencing public, private and blind rolls.
image

### DIFF
--- a/style/theme/_chat.scss
+++ b/style/theme/_chat.scss
@@ -53,15 +53,9 @@ body:not(.no-theme):not(.theme-light) #chat-popout,
         padding: 1re 0.5rem;
 
         &.whisper {
-            border-image: url(./ui/backgrounds/chat-message-border-private.webp) 21 repeat;
-            border-image-outset: 8px;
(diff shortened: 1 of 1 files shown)


style/theme/_chat.scss
```diff
@@ -52,6 +52,19 @@ body:not(.no-theme):not(.theme-light) #chat-popout,
...
+        &.whisper {
+            border-image: url(./ui/backgrounds/chat-message-border-private.webp) 21 repeat;
+            border-image-outset: 8px;
+            border-image-width: 12px;
+            border-image-slice: 12;
+            &.blind {
+                border-image: url(./ui/backgrounds/chat-message-border-blind.webp) 21 repeat;
+                border-image-outset: 8px;
+                border-image-width: 12px;
+                border-image-slice: 12;
```
moo-man, line 64: If you use `border-image-source` you don't need to redefine outset, width, and slice.

Issue #1661, opened 2023-07-01
[WFRP4e System] Add RAI houserule for Point Blank range


moo-man:
As explained by Andy Law, RAI meant to give 4 units (2 squares) as PBR (Point Blank Range) for all ranged weapons, instead of percentage based on their Range.

Could you add this as a house rule setting in the system? I can see `computeRangeBands()` function, but it has no hook so most likely I can't make it quick with a module.

Source: https://docs.google.com/document/d/1v37KH8Pa1BjTfnH4QT_N4_MGK3seKMvkvQKnWiJ4u3A/edit#heading=h.1qgs6qgua6b8 (Page 73, Chapter 5 -> Combat -> Ranged Combat -> Point Blank -> Q2 and Q3)

<br/>**From**: manufacture<br/>foundry: 11.304<br/>wfrp4e: 6.5.9<br/>wfrp4e-core: 3.1.4<br/>wfrp4e-up-in-arms: 3.0.6<br/>wfrp4e-wom: 3.0.5<br/>Active Modules: wfrp4e-unofficial-grimoire

moo-man:
This is not RAI, simply a house rule, and I generally don't include UFAQ

PR #2585, opened 2025-08-06
Static Point Blank Range Homebrew

marcin-jablonski:
Added homebrew setting for Point Blank range described in some interviews with creators as intended one:

- you can select if you want to use Static Point Blank Range in hombrew settings
- range by default is 4 yds (as proposed by creators), but can be configured in homebrew settings

### DIFF
--- a/src/model/item/weapon.js
+++ b/src/model/item/weapon.js
@@ -374,9 +374,14 @@ export class WeaponModel extends PropertiesMixin(EquippableItemModel) {
 
         let staticPointBlankRange = game.settings.get("wfrp4e", "homebrew").staticPointBlankRange;
 
-        let pointBlankRangeCutoff = staticPointBlankRange ? 
-            game.settings.get("wfrp4e", "homebrew").staticPointBlankRangeValue : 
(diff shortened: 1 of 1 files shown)


moo-man:
In the past I have not included UFAQ (https://github.com/moo-man/WFRP4e-FoundryVTT/issues/1661), but this is frequently asked for, so I'm not quite sure yet.

Forien:
I believe the homebrew should be "4yds or 10%, whichever is smaller"

Doesn't make sense for that homebrew to buff pistols and crossbow pistols even more.

marcin-jablonski:
> I believe the homebrew should be "4yds or 10%, whichever is smaller"
> 
> Doesn't make sense for that homebrew to buff pistols and crossbow pistols even more.

Right, makes sense. I'm just a bit too focused on elfbows here.

marcin-jablonski:
> In the past I have not included UFAQ (#1661), but this is frequently asked for, so I'm not quite sure yet.

Sure, up to you in the end. But it's here ;)

PR #2591, opened 2025-08-14
fix: Restore showing property description in chat

### LINKED ISSUE #2587: properties of objects are not revealed
Describe the bug
In the chat and in the character sheet, the description of the properties of items is not revealed. In the character sheet, only the Qualitis armor is not revealed.

Version Numbers
Foundry: 347
wfrp4e: 9.1.4
wfrp4e-core: 7.1.0
wfrp4e-archives1: 7.1.0
wfrp4e-archives2: 7.1.0
wfrp4e-archives3: 7.1.0
wfrp4e-up-in-arms: 7.1.0

Tiamanti:
Fixes #2587 
Restores the option to show item property on click
image

### DIFF
--- a/src/hooks/chat.js
+++ b/src/hooks/chat.js
@@ -79,11 +79,6 @@ export default function() {
       })
     })
 
-    // Post an Item Quality/Flaw
-    html.on("click", ".item-property", ev => {
-      WFRP_Utility.postProperty(ev.target.text)
-    })
-
     // if (app.getFlag("wfrp4e", "roleTests"))
     // {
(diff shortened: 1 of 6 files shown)


src/hooks/chat.js
```diff
@@ -78,6 +79,11 @@ export default function() {
       })
     })
 
+    // Post an Item Quality/Flaw
+    html.on("click", ".item-property", ev => {
+      WFRP_Utility.postProperty(ev.target.text)
+    })
+
```
moo-man, line 86: Is there a reason this was added back? I'm not accepting JQuery as I'm phasing it out with Foundry, but it doesn't seem to be needed in the first place.

PR #2603, opened 2025-09-04
Fix creating Fear test when actor has no Cool skill

marcin-jablonski:
When Actor doesn't have Cool Skill applying Fear (e.g. from `/fear` command) fails with error saying that Cool skill wasn't found.

Apparently this is because Fear effect item which is being duplicated from config has hardcoded "Cool" as test.

This fix checks for existence of Cool Skill and if needed updates duplicated Fear Effect to use Willpower.

### DIFF
--- a/src/documents/actor.js
+++ b/src/documents/actor.js
@@ -1102,6 +1102,11 @@ export default class ActorWFRP4e extends WarhammerActor
     fear.system.SL.target = value;
 
     foundry.utils.setProperty(fear, "flags.wfrp4e.fearName", name)
+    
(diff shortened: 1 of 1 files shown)


moo-man:
I would prefer to reimplement the previous behavior that handled this for all skills, instead of just Fear and Cool

marcin-jablonski:
Makes sense. I've had a look and it actually seems to be quite easy fix, something like [here](https://github.com/moo-man/WFRP4e-FoundryVTT/compare/master...marcin-jablonski:WFRP4e-FoundryVTT:fix-ext-test-when-no-skill?expand=1), so feel free to check it out and use if you want to.

PR #2604, opened 2025-09-07
localization

silentmark:
More localization strings, cleanup and order of en.json, for easier translations/change tracking.


src/system/config-wfrp4e.js
```diff
@@ -1891,7 +1892,7 @@ WFRP4E.PrepareSystemItems = function() {
                 scriptData: [
                     {
                         trigger: "manual",
-                        label: "@effect.name",
+                        label: game.i18n.localize("WFRP4E.ConditionName.Bleeding"),
```
moo-man, line 1895: I'm not sure why this is necessary if the effect name itself is localized?

silentmark: @effect.name doesn't work well with localization - for unknown reasons, when translation is enabled - it doesn't even keep the original (non-translated) value, but treats "@effect.name" as a literal and keeps it that way.

src/system/config-wfrp4e.js
```diff
@@ -2103,22 +2104,22 @@ WFRP4E.PrepareSystemItems = function() {
                 scriptData: [
                     {
                         trigger: "dialog",
-                        label: "Penalty to all Tests (@effect.name)",
+                        label: game.i18n.format("WFRP4E.ConditionEffects.PenaltyToAllTests", {effect: game.i18n.localize("WFRP4E.ConditionName.Stunned")}),
```
moo-man, line 2107: I feel like localizing script labels is a but much, most of them are hardly seen, and if it's necessary could just be replaced by the translation module?

silentmark: this would require replacement of whole object in the config, which i wanted to avoid in this case. And since there weren't so many - I wanted to keep it simple. but again, its your call - i can remove it.

silentmark: @moo-man  - please let me know, if you are in the end ok with this localization approach?

silentmark:
@moo-man - i've updated the PR for latest changes.

src/system/config-wfrp4e.js
moo-man: I'm going to say let's revert this one.

src/system/config-wfrp4e.js
moo-man: If this doesn't work with localizations then it should be investigated, as it would imply issues with other effects.

src/hooks/getSceneControlButtons.js
```diff
@@ -10,7 +10,7 @@ export default function() {
       button: true,
       icon: "fas fa-circle",
       name: "morrslieb",
-      title: game.canvas.scene?.getFlag("wfrp4e", "morrslieb") ? "Morrslieb - Currently On " : "Morrslieb - Currently Off",
+      title: game.canvas.scene?.getFlag("wfrp4e", "morrslieb") ? game.i18n.localize("Morrslieb - Currently On") : game.i18n.localize("Morrslieb - Currently Off"),
```
moo-man, line 13: This one is probably worth adding a key for

PR #2774, opened 2026-04-30
localizations

Txus5012:
A few localizations for the Extra Overcast options and Changes tab.


Txus5012:
Added localization for Rest & Recover.

Txus5012:
Conflict resolved.

Issue #1763, opened 2023-09-07
[Feature request] Repelling the Winds 💨


GregorHak:
### Is your feature request related to a problem? Please describe.
Repelling the Winds rules are not currently integrated (see Additional context for full rules description). Core and WoM describe these rules, also probably every group with spellcaster in party uses this rules as commonplace. 

### Describe the solution you'd like:
Split the Repelling the Winds rules into two separate, different scripts.
First **Repelling the Winds (armour)**
<details> 
  <summary>Click here to expand the rules snippet for armour👇</summary>
All Casting and Channelling Tests suffer a –1 SL penalty if you are dressed inappropriately for the Wind of Magic you are trying to attract, as determined by the GM. Specifically, metal and leather armour repel most of the Winds: metal is laden with the golden wind Chamon , while leather retains traces of the amber Ghur . As such, spellcasters wearing armour suffer –1 SL penalty to all Casting and Channelling Tests for every Armour Point on the location with the most armour. Casters with the Arcane Magic (Metal) Talent may wear metal armour without penalty; those with Arcane Magic (Beasts) Talent may ignore penalties from leather armour.
</details>

 three suggestions on how to do this:

1. **Hardcoded script**, as is the case for Wearing any Mail or Plate confers a Penalty of -10 Stealth. The potential downside is that it can affect every actor, not just wizard players.
2. **Effect script on the all Armour items types**. You will need to add this effect for each item separately, but the advantage is that you can remove/disable it as GM/Player at any time on the character sheet.
3. **System effect to choose from the drop-down list in the Effect tab on the character sheet**. See Additional context for the screen of full drop-down list if you don't understand what I mean. If you don't like the above two suggestions, please consider at least this one, some players take this penalty into account, and now must use their own script or rely on memory. 

Below is a script you might consider using. I don't know the original author, but the script have recently been heavily reworked by Perkuns on Moohammer discord. The script also takes into consideration the channeling skill test from the Skills (PC)/Main (NPC) tab.

```javascript
let isSkill = args.item.type === "skill";
let isCast = args.type === "cast";
let isChannel = args.type === "channelling";
let isChannelOrCastSkill = isSkill && args.item.name.includes("Channelling")
let table = this.actor.system.status.armour
let modifier = 0
if (!isChannelOrCastSkill && (isCast || isChannel)) {
    //loop for every location excluding shield location
    for (let i = 0; i < 6; i++) {
[...]

PR #2792, opened 2026-06-11
Spellcaster armour

Tiamanti:
Implements Repelling the Winds, CB p.237 rules.
Works with Archives 3 armour.
Allows turning this off as homebrew rule to preserve existing behaviour.
On by default.


GregorHak:
https://github.com/moo-man/WFRP4e-FoundryVTT/issues/1763

PR #2793, opened 2026-06-11
feat: stack quantity when purchasing duplicate items from market

Tiamanti:
When an actor buys a posted item they already own (same name and compendium source), increase the existing item's quantity instead of creating a duplicate. 
Adds MARKET.ItemAppended notification to distinguish the two cases.

### DIFF
--- a/src/sheets/actor/base.js
+++ b/src/sheets/actor/base.js
@@ -373,6 +373,35 @@ export default class BaseWFRP4eActorSheet extends WarhammerActorSheetV2
         }
       },
       {
+        name: "Buy More",
+        icon: '<i class="fas fa-shopping-cart"></i>',
+        condition: li => {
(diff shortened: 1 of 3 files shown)


src/sheets/actor/base.js
```diff
@@ -372,6 +372,35 @@ export default class BaseWFRP4eActorSheet extends WarhammerActorSheetV2
           document.postItem();
         }
       },
+      {
+        name: "Buy More",
```
Tiamanti, line 376: This was meant to be a branch based off this branch but I messed up and change is small so it might be fine.
It adds new option for Trappings right click menu called "Buy More" it works similar to "Post to Chat" but when available posts the compendium item instead of the player copy.
Useful for Arrows, Potions and such

PR #2804, opened 2026-06-20
calculate spell cost

Txus5012:
Fix for calculateSpellCost after the magic refactor.

Also added "undivided" to the chaos spell lores.
Note that the Enemy in Shadows spells have "Undivided" lore value, which should be changed to "undivided". Just remove the lore and add the Undivided lore again to fix it.

### DIFF
--- a/src/system/advancement.js
+++ b/src/system/advancement.js
@@ -226,33 +226,33 @@ export default class Advancement
     }
 
 
-    if (["slaanesh", "tzeentch", "nurgle", "undivided"].includes(spell.lore.value[0]))
+    if (["slaanesh", "tzeentch", "nurgle", "undivided"].includes(spell.lore.value))
       return 0
 
-    if (spell.lore.value[0] == "petty" || spell.lore.value[0] == game.i18n.localize("WFRP4E.MagicLores.petty"))
+    if (spell.lore.value == "petty" || spell.lore.value == game.i18n.localize("WFRP4E.MagicLores.petty"))
       bonus = actor.characteristics.wp.bonus
     else 
       bonus = actor.characteristics.int.bonus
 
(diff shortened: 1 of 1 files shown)


Txus5012:
Added the cost of High Magic Spells, which is the double of the arcane spells.
Similar to the EiS spells I mentioned above, in the High Elf Bundle some spells have "high" lore value and others have "High Magic". Again, removing the lore and adding it again resolves the issue.

I'm not sure what is the cost of the Elven Arcane Spells.

src/system/advancement.js
```diff
@@ -226,32 +226,33 @@ export default class Advancement
...
-      currentlyKnown = actor.itemTags["spell"].filter(i => i.lore.value == spell.lore.value && i.memorized.value).length;
+      currentlyKnown = actor.itemTags["spell"].filter(i => i.lore.value[0] == spell.lore.value[0] && i.memorized.value).length;
     }
-    else if (spell.lore.value == "petty" || spell.lore.value == game.i18n.localize("WFRP4E.MagicLores.petty"))
+    else if (spell.lore.value[0] == "petty" || spell.lore.value[0] == game.i18n.localize("WFRP4E.MagicLores.petty"))
     {
-      currentlyKnown = actor.itemTags["spell"].filter(i => i.lore.value == spell.lore.value).length;
+      currentlyKnown = actor.itemTags["spell"].filter(i => i.lore.value[0] == spell.lore.value[0]).length;
       if (currentlyKnown < bonus)
         return 0 // First WPB petty spells are free
```
moo-man: I feel like this would constitute adding a getter for the first (or primary?) lore because this is ridiculous lol

Txus5012: OK. I reverted the changes about lore values and kept only the changes for Undivided and High Magic costs.

moo-man: Why revert? I'm just saying this could be cleaned up with a getter

Txus5012: I prefer to leave it to you.

